PR #34, opened 2014-08-08
Removing subsort KItem ::= Type to improve parsing times.

radumereuta:
@dwightguth Please run the tests for this pull request.
I've made some optimizations that should improve the parsing times considerably, and I want to see if they are safe.


chathhorn:
I assume this is obsolete now.

PR #70, opened 2015-01-29
Upgrades and cleanup to semantics + reorganize test suite

dwightguth:
@chathhorn please review

I suggest you review this one commit at a time. You will have to check out the middle commit on your machine, but it's almost entirely renames so it shouldn't be that bad to review.


chathhorn:
I've been playing around with these changes and they look good to me. You might consider removing tests/Makefile if it's obsolete now, though. Also, it looks like there are a few warnings about undeclared or singleton variables that you might want to fix.

PR #71, opened 2015-02-02
Various fixes

chathhorn:
Mostly stuff that got lost in the port. @dwightguth please review.


semantics/language/execution/check-loc.k
```diff
@@ -70,112 +80,101 @@ module C-CHECK-LOC
...
-     //           "An object which has been modified is accessed through an expression based on a restrict-qualified pointer to a const-qualified type.",
-     //           "6.7.3.1:4"))
-     //           ~> checkRestrict'(basedOn(Base:SymBase, _))
-     //      ...</k>
-     //      <mem>... Base |-> object(T:Type, _, _) ...</mem>
-     //      requires isConstType(innerType(T))
-     //      [structural]
+     // Split out the prov to normalize the address.
+     rule checkRestrict(loc(Base:SymBase, Offset:Int))
+          => checkRestrict(loc(Base, Offset), .Set)
```
dwightguth: so, this isn't an action to be taken immediately, but just a thought: maybe we should declare the following rule as a macro?
```
rule loc(Base:SymBase, Offset:Int) => loc(Base, Offset, .Set)
```

This would potentially significantly simplify the semantics in places while still allowing us to match on the 2-argument form.

semantics/language/execution/expr/function-call.k
```diff
@@ -128,17 +127,13 @@ module C-EXPR-FUNCTION-CALL
           ) ...</call-stack>
           <control>
                <curr-tu> OldTu:K => Tu </curr-tu>
-               <curr-function> CurrFun:K </curr-function>
+               <curr-function> CurrFun:K => X </curr-function>
```
dwightguth: why did you do this? the effect of this behavior is to make it so that the function name and the code location are out of sync if an error occurs between the application of this rule and the entry to the block. That is why I originally made this modification.

tests/unit-fail/j041d-link1.ref
```diff
@@ -1,13 +0,0 @@
-
```
dwightguth: looks like this isn't the only .ref file that is now obsolete. Can you also delete the others?

dwightguth: oh never mind, looks like you are deleting them, git just did something really weird with the diff and thought they were renames. Okay, you can ignore this comment.

dwightguth:
Looks good, and yay for more passing tests. Just one thing I'd like to clear up before we merge this.

semantics/language/execution/expr/function-call.k
chathhorn: I'm not really sure what the best solution is, but probably for now the current code location should just be invalidated at the time of a call. Or maybe set to some special value to indicate that we're currently binding params for a call to function x from location blah. Or maybe the error reporting mechanism could be made smarter in order to detect this case.

Setting curr-function like that might keep it consistent with the current code loc, but it's inconsistent with the current TU and the current call stack (also it breaks some restrict checks).

PR #72, opened 2015-02-03
canonicalize FieldInfo to be stored in only one place in the semantics

dwightguth:
@chathhorn please review


chathhorn:
Looks good!

PR #73, opened 2015-02-10
various minor fixes to c semantics

dwightguth:
@chathhorn please review

I had to add the .h file to the distribution because technically the version that comes with glibc is invalid C code: it declares an array of length 0 instead of a flexible array member.


chathhorn:
Looks good!

PR #76, opened 2015-02-19
Fix regressions

chathhorn:
@dwightguth


semantics/language/common/compat.k
```diff
@@ -240,4 +242,6 @@ module COMPAT
      rule dotK => .K [structural]
      rule lookup(Key |-> Val:K _:Map, Key:K) => Val
 
+     rule isKItem(K:KItem) => true
```
andreistefanescu: Can you try without these two rules. I think it might work based on default sort handling.

dwightguth: Tested, and it does not actually rewrite to true unless these rules exist.

PR #77, opened 2015-02-22
Last of the regressions

chathhorn:
Should finally all be passing again. @dwightguth


tests/unit-fail/pta-field-2.ref
```diff
@@ -0,0 +1,13 @@
+
```
dwightguth: I am looking at this program and it doesn't look correct to me that it generates an out-of-bounds error. The line of code in question is taking a pointer to an int* and subtracting 1 from it. But the memory being pointed to is the second int* in a struct. So unless I am reading this wrong, it should actually be correctly pointing to the first member of the struct. Can you explain why you think this program should be undefined?

chathhorn: It's a pointer created by & applied to a scalar, which is treated as a pointer into an array of a single element according to the standard. That it also happens to be a field in a struct isn't relevant. Any pointer arithmetic resulting in a pointer outside the bounds of an array (+ 1 past the end) is undefined.

dwightguth: Okay, cool. Didn't know that. Thanks!

PR #84, opened 2015-03-10
Declare sorts for different kinds of  types + minor fixes

dwightguth:
@chathhorn please review

This might actually make the Jenkins build a little bit slower, for a couple reasons (one is that the parse table takes longer to generate, the other is that it takes longer to compute the sorts of terms in the java backend) but it is made up for by the considerably decreased amount of time spent in side conditions during rewriting.

I also made a few other minor tweaks to get my C code to run,, and run in a reasonable time frame.

### DIFF
--- a/semantics/language/common/binding.k
+++ b/semantics/language/common/binding.k
@@ -146,11 +146,10 @@ module C-BINDING
                andBool representable(V, argPromoteType(T'))
                andBool representable(V,
                     correspondingUnsignedType(argPromoteType(T)))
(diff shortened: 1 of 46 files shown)


dwightguth:
@chathhorn don't merge this yet actually because I realized the other day it causes some tests to fail that I need to fix before we try to merge this. But you're welcome to review what there is so far.

dwightguth:
Okay, I think I have this passing all the tests now. @chathhorn , can you please review this?

chathhorn:
This all looks good!

PR #85, opened 2015-03-15
Pragmas, compound lits, etc.

chathhorn:
@dwightguth

### DIFF
--- a/semantics/language/execution/compound-literal.k
+++ b/semantics/language/execution/compound-literal.k
@@ -57,10 +57,23 @@ module C-COMPOUND-LITERAL
                ~> X
           ...</k>
           <compound-lit-map> M:Map => M[X <- Base] </compound-lit-map>
+          <local-addresses>... .Set => SetItem(Base) </local-addresses>
           <next-loc> Base:SymBase => linc(Base) </next-loc>
           requires notBool (X in keys(M))
           [structural]
-
+     rule <k> handleCompoundLiteral(initValue(X:CId, T:Type, Init:K))
+               => allocObject(Base', T)
+               ~> addToEnv(X, Base')
+               ~> giveType(X, T)
+               ~> allowInit(Init)
+               ~> sequencePoint
+               ~> X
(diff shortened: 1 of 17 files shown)


semantics/language/execution/compound-literal.k
```diff
@@ -57,10 +57,23 @@ module C-COMPOUND-LITERAL
                ~> X
           ...</k>
           <compound-lit-map> M:Map => M[X <- Base] </compound-lit-map>
+          <local-addresses>... .Set => SetItem(Base) </local-addresses>
```
dwightguth: can you explain to me what is going on with this change here?

semantics/language/translation/compound-literal.k
```diff
@@ -58,15 +58,13 @@ module C-COMPOUND-LITERAL
...
-               ~> Init
+               ~> allowInit(Init)
                ~> X
           ...</k>
           <curr-tu> Tu:String </curr-tu>
           <tu-id> Tu </tu-id>
           <next-static-loc> Base:SymBase => linc(Base) </next-static-loc>
-          <initializing> I:Int </initializing>
-          requires I >Int 0
+          <block-history> .List </block-history>
```
dwightguth: and here?

semantics/language/execution/compound-literal.k
chathhorn: This adds the newly allocated address to the set of local addresses so it'll be cleaned up/invalidated when the compound literal object goes out of scope. Compound literals at block scope are supposed to have a lifetime of the block.

semantics/language/translation/compound-literal.k
chathhorn: This prevents block-scope compound literals from being allocated at translation-time. Global-scope literals can be allocated, as they're supposed to have static storage class and a lifetime for the whole program.

dwightguth:
Okay. If you fix the conflicts, you can merge this.

PR #90, opened 2015-04-09
first batch of small changes

andreistefanescu:


semantics/language/common/symloc.k
```diff
@@ -42,9 +42,10 @@ module C-SYMLOC-SYNTAX
 
 
      // Base, byte offset, bit offset.
-     syntax SymLoc ::= loc(SymBase, Int, Set)
      syntax SymLoc ::= loc(SymBase, Int)
-     syntax SymLoc ::= "NullPointer"
+     syntax SymLoc ::= HeapSymLoc
```
ellisonch: There is no heap in C :)

dwightguth: I'd go with CanonicalSymLoc

semantics/language/common/typing/misc.k
```diff
@@ -331,6 +331,7 @@ module C-TYPING-MISC
      rule getReadFrom'(K:K, S:Set) => getReadFrom'(.K, S)
           requires getKLabel(K) =/=KLabel 'readFrom
                andBool (K =/=K .K)
+     rule getReadFrom'(.K, .Set) => NullPointer
```
dwightguth: this does not seem correct to me... I believe the motivation is supposed to be that this function is only called in contexts where hasReadFrom has already evaluated to true. Maybe we should create some kind of special bottom value if you need the function to be total, but NullPointer is a valid SymLoc, which is not where this particular value was actually read from.

semantics/language/common/symloc.k
andreistefanescu: Sounds good.

andreistefanescu:
Jenkins: test this please

andreistefanescu:
Jenkins: test this please

dwightguth:
Jenkins: test this please

andreistefanescu:
@dwightguth please review

PR #92, opened 2015-04-10
remove hacky version checking

dwightguth:
If you have a compiler that supports -std=c11 but doesn't actually
support c11 sufficiently, on your own head be it.

@andreistefanescu can you try this and if it fixes your problem, merge?

### DIFF
--- a/scripts/kcc
+++ b/scripts/kcc
@@ -354,7 +354,7 @@ sub preprocess {
      my ($output, $inputFile) = (@_);
 
      my $directoryname = dirname($inputFile);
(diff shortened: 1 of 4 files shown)


scripts/kcc
```diff
@@ -357,7 +357,7 @@ sub preprocess {
      system("which gcc-4.9 > /dev/null 2>&1");
```
andreistefanescu: should this be gcc as well (not gcc-4.9)?

dwightguth: yes thanks

PR #93, opened 2015-04-15
add code to C semantics for annotation generation

dwightguth:
@andreistefanescu @chathhorn please review

### DIFF
--- a/semantics/language/translation/init.k
+++ b/semantics/language/translation/init.k
@@ -1,4 +1,5 @@
 module C-TRANSLATION-INIT-SYNTAX
+     imports KCELLS
      syntax KItem ::= load(Bag, Bag, Bag, Bag)
      syntax KItem ::= "link" | "cleanup"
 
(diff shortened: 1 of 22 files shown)


semantics/language/translation/init.k
```diff
@@ -260,5 +261,26 @@ module C-TRANSLATION-INIT
      syntax KItem ::= "reportSuccess"
      rule <k> reportSuccess => .K ...</k>
           <result-value> _ => 0 </result-value>
+          <generated-annotations> .List </generated-annotations>
+     rule <k> reportSuccess => writeSpecModule(#open(S +String ".spec.k")) ...</k>
```
andreistefanescu: maybe ```S +String "_spec.k"```?

andreistefanescu:
lgtm

andreistefanescu:
lgtm

semantics/language/translation/init.k
dwightguth: we could do that, but that will lead to files called foo.c_spec.k or foo.h_spec.k, which I am inclined to think looks worse than foo.c.spec.k or foo.h.spec.k.

dwightguth:
Jenkins: test this please

PR #95, opened 2015-04-22
moved effective types inside object

andreistefanescu:


andreistefanescu:
@dwightguth @chathhorn please review

dwightguth:
looks good to me, but I'm not hugely familiar with the effective type code. @chathhorn , can you weigh in? Essentially what this diff is doing is just moving the effective-types cell into the <mem> cell.

chathhorn:
Looks fine to me.

PR #96, opened 2015-04-27
apply chucky's patch to remove profiling

dwightguth:
reviewing myself since it's Chucky's code originally


dwightguth:
lgtm. Will merge if tests pass

PR #98, opened 2015-04-28
fixed cleanup before return

andreistefanescu:


andreistefanescu:
@dwightguth @chathhorn please review

dwightguth:
I'm kinda surprised this passes all the tests and doesn't degrade the quality of the error messages, but since it does, lgtm.

andreistefanescu:
@dwightguth why are you surprised?

dwightguth:
well I would have thought that deleting the objects at the end of the block would have consequences for code that is trying to determine whether an object is currently live.

andreistefanescu:
the previous code was also deleting the objects, but I think is was less clear what was happening.

dwightguth:
Oh I see. I misread the change then. yeah, this seems fine

chathhorn:
Looks fine.

PR #102, opened 2015-05-08
added extra condition to prefent rule race

andreistefanescu:


andreistefanescu:
@chathhorn please review

chathhorn:
Looks good.

PR #105, opened 2015-05-28
Refactoring to streamline execution, cleaner version

chathhorn:
Reorganized the commits.


dwightguth:
For each of the files that was deleted without is rules being moved into a new location, can you explain to me how the equivalent piece of the semantics is now being handled?

semantics/language/common/typing/misc.k
```diff
@@ -237,6 +239,14 @@ module C-TYPING-MISC
 
      rule addModifiers(Mods':Set, t(Mods:Set, T:SimpleType))
           => t(Mods' Mods, T)
+     // TODO(chathhorn): !?!!
```
dwightguth: um, why is this happening?

semantics/language/execution/io-buffered.k
```diff
@@ -173,6 +173,14 @@ module C-IO-BUFFERED
 
      rule updateObject(object(T:K, Len:Int, Bytes:Map, Types:Map), Offset:Int, V:K)
           => object(T, Len, Bytes[Offset <- V], Types)
+     // !!?!
+     rule updateObject(object(T:K, Len:Int, Bytes:Map, .Map), Offset:Int, V:K)
```
dwightguth: same here

semantics/language/execution/stmt/return.k
```diff
@@ -51,8 +52,8 @@ module C-STMT-RETURN
                     | "Return''" "(" K ")"
      context Return'(HOLE => reval(HOLE)) [result(RValue)]
 
-     /* recursively exist each Block wrapped in BlockStatement */
-     rule <k> (.K => exitRestrictScope(BlockNum) ~> deleteLocals ~> popBlock)
+     /* recursively exits each Block wrapped in BlockStatement */
+     rule <k> (.K => exitRestrictScope(BlockNum) ~> popBlock)
```
dwightguth: I seem to recall Andrei added this because he needed it. Why was it removed? also, small items like this in a large diff are quite critical but very hard to notice, and I can't necessarily guarantee that I will notice all of them. Can you remember any other cases in this diff where you made changes that were not immediately related to the task you were working on? it could be quite important.

semantics/language/translation/configuration.k
```diff
@@ -93,25 +86,31 @@ module C-CONFIGURATION
...
-               UnknownCabsLoc
-          </curr-program-loc>
-
-          // Toggle to allow writes and other special treatment for
-          // initializers.
-          <initializing> 0 </initializing>
-
-          <control-details> // TODO(chathhorn): dumb name
+     <elab-calc>
+          <elaborator multiplicity="*">
```
dwightguth: multiplicity * cells have a potentially significant impact on performance. Are you sure that this one is necessary, or can we get away with sequentially analyzing each elaboration task without resorting to a * cell?

semantics/language/translation/decl/global.k
```diff
@@ -437,19 +444,19 @@ module C-DECL-GLOBAL
      // differ in the case of declarations without initializers. A naked
      // declaration without an initializer constitutes a tentative definition,
      // but an extern declaration without an initializer does not.
-     syntax KItem ::= "declareExternalVariable'" "(" K "," DeclState "," Linkage "," Storage ")"
-          [strict(1)]
+     syntax KItem ::= "declareExternalVariable'" "(" Bool "," K "," DeclState "," Linkage "," Storage ")"
```
dwightguth: what is this added field doing?

semantics/language/execution/stmt/return.k
chathhorn: I did refactoring all over the place. If this re-introduces Andrei's issue (it'd be nice if there was a regression test, of course) we can re-fix it, but how blocks were dealt with was all refactored (Block gets expanded to pushBlock ~> ... ~> popBlock during translation and popBlock => deleteLocals).

semantics/language/translation/configuration.k
chathhorn: Elaboration of function bodies could possibly be done in parallel (by saving the global state before each function), which I'd think would be desirable. But it could be done with like a stack too I suppose.

semantics/language/translation/decl/global.k
chathhorn: The initializer coming back empty after calculating it doesn't necessarily mean it was an empty initializer to begin with because it's the "frozen" initializer that's returned. If there was nothing to freeze (it could all be run at translation-time) it'll just be .K.

semantics/language/common/dynamic.k
```diff
@@ -15,6 +15,13 @@ module C-DYNAMIC-SYNTAX
 
      syntax LValue ::= lv(SymLoc, Type)
 
+     syntax KItem ::= le(K, Type)
```
andreistefanescu: A more specific sort than ```K``` would make the semantics more readable.

semantics/language/common/dynamic.k
```diff
@@ -15,6 +15,13 @@ module C-DYNAMIC-SYNTAX
...
+     syntax Hold ::= te(K, Type)
+     syntax LHold ::= ncle(K, Type)
+     syntax CompoundExpression ::= compoundExp(K)
+
+     syntax KItem ::= stripHold(K) [function]
```
andreistefanescu: See above.

semantics/language/common/dynamic.k
```diff
@@ -23,7 +30,7 @@ module C-DYNAMIC-SYNTAX
      syntax KItem ::= toRVal(K)
 
      syntax Type ::= type(K) [function]
-     syntax KItem ::= value(RValue) [function]
+     syntax KItem ::= value(K) [function]
```
andreistefanescu: See above.

semantics/language/execution/ltlmc.k
```diff
@@ -93,17 +92,15 @@ module C-LTLMC
...
           => ltlRVal(B, * (L + R))
-     */
 
      // I don't really think & is necessary.
      // rule ltlRVal(B:Bag, & P:CProp)
      //      => ltlLVal(B, P)
 
      // "strictness"
-     /*TODO(liyili2): cannot handle this cases
+     /*TODO(liyili2): cannot handle this cases */
```
andreistefanescu: Is this still the case?

semantics/language/translation/configuration.k
andreistefanescu: This is only in the translation semantics, so performance is less of an issue.

andreistefanescu:
Chris, this is awesome! My one suggestion is to limit the use of the ```K``` sort as much as possible, as use specific sorts instead, in order to improve readability.

semantics/language/translation/configuration.k
dwightguth: Performance is less of an issue in the translation semantics but still
significant insofar as it can take half an hour to an hour in the current
rewrite engine to compile a large real code project. But we can leave it
this way for now and run experiments with changing it if things seem so
slow.

dwightguth:
looks really great! I am okay with merging this if @andreistefanescu is okay and then I can create a minor pull request for the multiplicity issue. That way @chathhorn doesn't have to spend more of his limited time on this diff.

PR #107, opened 2015-06-25
fix nontermination bug in compositeFunctionType

dwightguth:
@andreistefanescu please review

### DIFF
--- a/semantics/language/common/typing/compatibility.k
+++ b/semantics/language/common/typing/compatibility.k
@@ -352,8 +352,10 @@ module C-TYPING-COMPATIBILITY
           => t(Mods, functionType(T, P'))
      rule compositeFunctionType(T:Type, Mods:Set, .K, .List, P:List, _)
           => compositeFunctionType(T, Mods, .K, .List, .List, P)
+          requires P =/=K .List
      rule compositeFunctionType(T:Type, Mods:Set, .K, P:List, .List, _)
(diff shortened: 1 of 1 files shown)


semantics/language/common/typing/compatibility.k
```diff
@@ -352,8 +352,10 @@ module C-TYPING-COMPATIBILITY
           => t(Mods, functionType(T, P'))
      rule compositeFunctionType(T:Type, Mods:Set, .K, .List, P:List, _)
```
grosu: Isn't it nicer/faster to write such rules using local rewrites, like:

    rule compositeFunctionType(T:Type, Mods:Set, .K, .List, `P:List => .List`, `_ => P`)
    requires P =/=K .List

dwightguth: I think the way this was is left over from a limitation of the Java rewrite engine which is no longer present, wherein such rewrites in function rules were not allowed. In any event, I'm not sure it's worth the effort to change at this point.

andreistefanescu:
lgtm

semantics/language/common/typing/compatibility.k
grosu: Not at this point, but it may be worth beautifying a bit the C semantics at some later point.  Maybe have somebody else (an intern?) do it.  The C semantics is one of the oldest semantics we have, so no wonder that it may not take advantage of the latest developments.  So ok to merge.

PR #114, opened 2015-08-11
Update C semantics to use OCAML backend + KORE

dwightguth:
@bmmoore FYI, feel free to take a look at the makefile changes

@andreistefanescu @chathhorn please review

Let me know if you want me to refactor the history into any particular form to make it easier to read.


Makefile
```diff
@@ -16,11 +16,11 @@ FILES_TO_DIST = \
 	$(SCRIPTS_DIR)/program-runner \
 	$(PARSER_DIR)/cparser \
 
-.PHONY: default check-vars semantics clean fast semantics-fast $(DIST_DIR) $(SEMANTICS_DIR)/settings.k $(SEMANTICS_DIR)/extensions.k test-build
+.PHONY: default check-vars semantics clean fast translation-semantics execution-semantics $(DIST_DIR) $(SEMANTICS_DIR)/settings.k $(SEMANTICS_DIR)/extensions-common.k $(SEMANTICS_DIR)/extensions-translation.k $(SEMANTICS_DIR)/extensions-execution.k test-build pass fail fail-compile
 
 default: dist
 
-fast: $(DIST_DIR)/lib/libc.so $(DIST_DIR)/c11-kompiled/c11-kompiled/context.bin
+fast: $(DIST_DIR)/lib/libc.so $(DIST_DIR)/c11-kompiled/c11-kompiled/def.cmx
```
bmmoore: You should pull out an EXTENSION variable in this makefile too.

libc/semantics/settings.k
```diff
@@ -40,13 +39,15 @@ module C-SETTINGS
...
      rule cfg:alignofDouble => 8
      rule cfg:alignofLongDouble => 16
-     rule cfg:alignofEnum => cfg:alignofInt
      rule cfg:alignofPointer => 4
      // Max of the above.
      rule cfg:alignofMalloc => 16
 
-     rule cfg:intToPointer(_, _) => trap
+     rule cfg:intToPointer(0, _) => NullPointer
+     rule cfg:intToPointer(_, _) => trap [owise]
```
andreistefanescu: does this mean you cannot cast a pointer into an integer and back into the same pointer?

semantics/language/common/expr/eval.k
```diff
@@ -41,19 +43,19 @@ module C-COMMON-EXPR-EVAL
      }*/
 
      // Lvalue conversion.
-     rule reval(nclv(Loc:SymLoc, T:Type)
+     rule reval(nclv(Loc::SymLoc, T::Type)
```
andreistefanescu: why ```::```?

semantics/language/common/syntax.k
```diff
@@ -4,10 +4,11 @@ module C-SYNTAX
      imports INT-SYNTAX
      imports BOOL-SYNTAX
      imports LIST
-     imports KORE-COMPAT
 
      imports C-DYNAMIC-SYNTAX
 
+     syntax KBott ::= "(" K ")" [bracket]
```
andreistefanescu: what's with this?

semantics/language/common/typing/misc.k
```diff
@@ -19,73 +19,85 @@ module C-TYPING-MISC
...
           => t(S, long-long-int)
+     rule correspondingSignedType(t(S:Set, bitfieldType(T:Type, I:Int)))
+          => t(S, bitfieldType(correspondingSignedType(T), I))
+     rule correspondingSignedType(t(S:Set, enumType(_))) 
+          => correspondingSignedType(t(S, cfg:enumAlias))
 
      rule min(t(_, bitfieldType(_:UnsignedIntegerType, _))) => 0
      rule max(t(_, bitfieldType(_:UnsignedIntegerType, N:Int)))
-          => (2 ^Int absInt(N)) -Int 1
+          => (1 <<Int absInt(N)) -Int 1
```
andreistefanescu: any particular reason for this change?

semantics/language/execution/check-loc.k
```diff
@@ -85,23 +85,23 @@ module C-CHECK-LOC
                (.Set => SetItem(basedOn(Base, Tag))) _)
           ...</k>
           <activeBlocks>... Tag |-> activity(_, Modified:Map) ...</activeBlocks>
-          requires Loc in keys(Modified)
+          requires Loc in_keys(Modified)
           [structural]
      rule <k> filterUnmodified(Loc:SymLoc, (basedOn(_, Tag:Scope) => .K), _, _)
           ...</k>
           <activeBlocks>... Tag |-> activity(_, Modified:Map) ...</activeBlocks>
-          requires notBool (Loc in keys(Modified))
```
andreistefanescu: what was wrong with the old ```in keys```?

semantics/language/execution/io-buffered.k
```diff
@@ -102,217 +88,188 @@ module C-IO-BUFFERED
...
                // TODO(chathhorn): do we really need to check this stuff every
                // byte?
-               => assert(notBool (loc(Base, Offset) in Locs),
-                    errorUnseqEffects)
-               ~> assert(Offset <Int Len, errorReadOutOfBounds)
-               ~> assert((notBool (Volatile in getQualsAtOffset(ObjT, Offset)))
-                    orBool isVolatileType(T),
-                    errorNonVolatileLVal)
-               ~> tv(V, t(.Set, no-type))
+               => #if notBool (loc(Base, Offset) in Locs) #then .K #else
```
andreistefanescu: why these changes?

dwightguth:
I need to close this pull request and reopen it because I messed up the merge. But I will address everyone's comments on this pull request as well.

libc/semantics/settings.k
dwightguth: technically the C standard does not guarantee the existence of the intptr_t and uintptr_t types, which are the only integer types guaranteed to be large enough to contain a pointer address and represent the same address when casted back. This implementation of C does not contain these types; the one that is in x86_64-linux-gcc-glibc does however. Essentially, settings.k contains implementation-defined behavior, and the behavior when an integer is cast to a pointer and back is implementation-defined.

semantics/language/common/expr/eval.k
dwightguth: faster

semantics/language/common/syntax.k
dwightguth: in order to avoid changing hundreds and hundreds of lines of code in the C semantics, I introduced legacy brackets with this production so that users don't have to say ``` `` ``` for brackets.

semantics/language/common/typing/misc.k
dwightguth: slightly faster

semantics/language/execution/check-loc.k
dwightguth: OCAML doesn't compute key sets in constant time, so I made a in_keys operator that will be in constant time.

semantics/language/execution/io-buffered.k
dwightguth: also faster.

semantics/language/common/expr/eval.k
andreistefanescu: I don't like this change. It may be faster, but it works against the type system. Next thing we know we have stuff other than ```SymLoc``` as the first argument because it type checks.

semantics/language/execution/io-buffered.k
andreistefanescu: Unless the performance penalty is 2x, I would revert back. The point is to have decent performance with the OCAML backend, not to sacrifice readability for the sake of performance. This is essentially manual inlining.

semantics/language/common/expr/eval.k
dwightguth: as a matter of fact, putting something other than a SymLoc as the first argument of an nclv will *not* typecheck unless the user explicitly casts a term placed in that position. In practice as long as you don't do this you will get an error if you make a mistake. I forgo the checks in a number of places in this diff, but all of them represent cases where my analysis indicates that the type system is guaranteed to not be violated.

The performance improvements associated with the substantial reductions in computing type information present in this diff are sufficiently significant that I don't really have any intention of compromising on this point, but if you want to propose some other kind of mechanism by which we can cut down on time spent computing sort information in critical rules of the semantics, I am willing to listen.

semantics/language/execution/io-buffered.k
dwightguth: I am well aware that that's what it is but this is the absolute most performance-critical section of the entire semantics. The point is to have decent performance, period, and having exactly one rule apply per byte read/written significantly cuts down on the total number of rewrites in most applications.

andreistefanescu: The point is not to have decent performance and period, because gcc has much better performance. The point is to have a semantics which is readable, easily maintainable, and has decent performance. I wasted months using the C semantics for verification as opposed to weeks for Java and JavaScript because of how the semantics is written. Refactoring is already painful enough.

Revert the manual inlining. If you feel so strongly about it, implement a inlining pass.

semantics/language/common/expr/eval.k
andreistefanescu: This is not just a matter of performance. If you work with klabels, you can easily violate the signature. Also for other backends may work better with the sort information. You are skewing the C semantics towards the OCaml backend. How would you feel if I started to change the semantics to make the verification faster or Brandon to make the Coq backend work better?

If the performance is so important, I suggest trying to cache the predicates, use OCaml datatypes, or otherwise accept this as a limitation of the OCaml backend.

semantics/language/execution/io-buffered.k
grosu: Can't we use a macro, which would do the inlining statically?  Would that solve the problem?

semantics/language/common/expr/eval.k
dwightguth: So Grigore reminded me that we have strict compilation on by default. I did some preliminary testing and it looks like if you don't use the --non-strict option, sort checks should get automatically inferred in cases like this, even when the user types `::` instead of `:`. Thus, you should be able to simply disable that flag and get the original behavior, which seems like a sufficient compromise if you are concerned about backends where increased sort information is desirable.

semantics/language/execution/io-buffered.k
grosu: At first sight, these rules look rather ugly and repetitive, both before and after @dwightguth 's changes.  Although the side conditions will be evaluated only after the matching takes place, and probably once that happens the side conditions hold with high probability, it still looks ugly and potentially inefficient to repeat the same side condition Size > 0 in each rule.  Maybe you guys can find a different way to write them all together, which would satisfy you both?

Orthogonally to that, the issue of writing rules like
```
1:
rule <k> readByte(stuff) => .K ...</k> <mem> sideEffects </mem> 
rule readBytes(N) => readByte(stuff) ~> readBytes(N - 1)
rule readBytes(0) => .K
```
versus
```
2:
rule <k> readBytes(N) => readBytes(N - 1) ...</k> <mem> sideEffects </mem>
rule readBytes(0) => .K
```
remains.

The benefit of 1 is that the rules look more "functional", or "inductive", but more rewrite steps are generated.
The benefit of 2 is performance of execution and fewer rules, but the rules are harder to reason with.

This is reminiscent of forwards recursive (1) versus tail recursive (2) functions in functional programming.  Although it is harder to reason about programs which are written in a tail recursive fashion, the performance benefits are often non-negligible and people prefer them over the ordinary recursive variants.  So in the end it boils down as a matter of subjective taste and preference.  Speaking of which, I would have probably preferred 2, too, because it gives me an opportunity to show off K's compact notation a bit more:
```
3:
rule <k> readBytes(N => N - 1) ...</k> <mem> sideEffects </mem>
rule readBytes(0) => .K
```

@andreistefanescu , how much does this affect you with your proofs?  If it is more a matter of taste and principle rather than technical, I would go with 2 or 3, because I think we will need to refactor the C semantics eventually anyway, and then we may rethink some of these issues, but at least, until then, we have a faster undefinedness checker.

andreistefanescu: @grosu my issue is with manual inlining, not with tail recursive vs. forward recursive. Manual inlining is a horrible software engineering practice, regardless of language. If @dwightguth replaces them with macros, I have no complaints.

PR #117, opened 2015-08-18
remove identifiers from environment that are not used in that TU

dwightguth:
**please review**

removes unused identifiers from genv and gtypes cells, plus removes all references to enumeration constants subsequent to translation phase.


dwightguth:
@chathhorn please review

chathhorn:
Looks good!

PR #120, opened 2015-09-02
Add many more error recovery cases + optimize KCC for RV-Match

dwightguth:
@chathhorn @andreistefanescu please review


dwightguth:
Jenkins: test this please

dwightguth:
@chathhorn @andreistefanescu please review

andreistefanescu:
Looks good. Do you have any plans to make error recovery more modular, so the user can specify which type of errors to recover from?

dwightguth:
I don't specifically have plans to do that myself; I think ultimately we may want that type of behavior, however. Maybe have Chris work on it at the same time as he works on the separation of the positive and negative semantics?

andreistefanescu:
I don't have any designs in this direction now. We can talk about it when the need arises.

chathhorn:
Looks good!

PR #122, opened 2015-09-29
Update README.md

chathhorn:
Added PLDI reference, removed LTL stuff.


grosu:
Thanks, @chathhorn !  Since we are at this topic, could you please revise the entire README and INSTALL, and be happy with what is there?  Then once @dwightguth does the same, we are done with it.

PR #128, opened 2015-10-09
configurable preprocessor in profile

dwightguth:
@chathhorn please review


tests/unit-fail-compilation/j060d.ref
```diff
@@ -1,7 +1,7 @@
 File: j060d.c
 Line: 2
 Error: TDG2
-Description: declaration for f appears with multiple storage class specifiers.
-Type: Constraint violation.
-See also: C11 sec. 6.7.1:2
+Description: declaration for f appears with invliad storage class specifiers.
```
chathhorn: Spelling.

scripts/kcc
```diff
@@ -41,6 +41,16 @@ if ($cygwin) {
...
+
+my $preprocessor;
+{
+  local $/=undef;
+  open FILE, "$profileDirectory/pp" or die "Couldn't find current profile: run kcc-switch";
+  $preprocessor = <FILE>;
+  close FILE;
+  chop($preprocessor);
+}
+
```
chathhorn: Did you consider maybe just making $profileDirectory/pp the executable preprocessor (a script or even a link)? 

Looks good to me, though.

dwightguth: I will probably do that eventually, honestly, but until we have profiles that use non-gcc preprocessors, it will be less code duplication this way. Eventually I want to have profiles be configured from individual components, so that that code duplication wouldn't happen, but that's more work to deal with right now.

PR #136, opened 2015-10-21
Fixes #132.

### LINKED ISSUE #132: kcc on structs with no named members
Consider the program:
struct a{
    int : 5
};

int main() {
    printf("Execution");
}
Compiling this file with kcc, and then running it works fine, and "Execution" is printed to the screen. However, according C11 sec 6.7.2.1,  "A structure or union is defined as containing no named members" is undefined. Shouldn't kcc also be reporting undefined  behavior here?

chathhorn:
Fix for the issue with fieldless structs.


dwightguth:
can you add a test?

Issue #135, opened 2015-10-20
test-build failing


bmmoore:
The test-build target fails trying to run the test program because `-` is apparently no longer supported as the name of the input file:

    make[1]: Leaving directory '/home/brandon/code/k/c-semantics/semantics'
    Testing kcc...
    printf "#include <stdio.h>\nint main(void) {printf(\"x\"); return 42;}\n" | dist/kcc - -o dist/testProgram.compiled
    Unsupported option - at dist/kcc line 333.
    Makefile:65: recipe for target 'test-build' failed
    make: *** [test-build] Error 255

dwightguth:
Fixed #130

PR #137, opened 2015-10-21
Use file instead of pipe in test-build

bmmoore:
A quick workaround for #135, which makes an intermediate file so test-build doesn't rely on `-` working as a file name.


dwightguth:
I'm inclined not to bother with this because my next diff intends to fix the underlying cause of the issue.

bmmoore:
Fixing kcc would be good.

PR #142, opened 2015-10-27
update INSTALL with new instructions

dwightguth:
@grosu @chathhorn please review


INSTALL.md
```diff
@@ -38,33 +39,44 @@ $ sudo apt-get install build-essential
     - XML::LibXML::Reader
     - String::Escape
 
-To install these modules using cpan:
+You can also insatll them using apt-get with
```
grosu: spell-check

grosu:
lgtm.  i'm assuming the native library executions is a separate module, right?

dwightguth:
Yeah the native library code is built using a separate makefile that we keep in the proprietary repo. Since we don't intend anyone outside the company to build that, it's not as important to have install instructions. I can just let people know what they have to do.

grosu:
OK.  So what happens now when people build the C semantics themselves without the native library and they attempt to execute a program that makes use of libraries?  Do they get an actual meaningful error message, or just a parsing error?

dwightguth:
They will most likely get an error when they link the program saying that no external definition was found for one of the symbols that they used which does not have semantics for it.

grosu:
I think it would be useful and in RV.com's favor to give an error message saying that in order to execute library code they need to get a license from RV.com.

dwightguth:
Not really something that's easy to do though. It's the exact same error whether they are linking against a function in the standard library that wasn't defined because they don't have the plugin, or if they are linking against a function that was declared in a header file but the user simply forgot to define it somewhere in their program. The latter is a user error and it would be inappropriate to show the type of error you are talking about.

I am open to suggestions for how to deal with this, though... We could maybe maintain a list of functions that should be defined by libc and are therefore something that the plugin would take care of...  but it would be a really long list and I'm not entirely sure how to obtain it easily.

grosu:
How about giving more of a suggestion than error message, saying something like (but improve, please):

No semantic definition was found for symbol xyz.  This typically means that you either declared xyz in a header file but forgot to define it somewhere in your program (in which case you would also get an error when executing your program after compilation with a standard C compiler), or that xyz is part of some library which currently has no semantics given.  If the latter, contact runtimeverification.com/support, because they provide a module extending the C semantics with library support as part of their commercial RV-Match system.

dwightguth:
So, maybe I should have been a little more precise. The error I described is the error you would get if you use a function declared in a header file that we never gave semantics to. There is also another type of error that you might see if you are using library functions in your program, which would be something more along the lines of "trying to look up identifier foo, but no such identifier is in scope". This occurs when you are using a function that is in the library in the standard, but the current profile of the C semantics does not include a prototype of it in the header file. You would also see this error if you forgot to include the appropriate header file.

Now we could probably modify the error messages in these two cases to do something like you describe, but then we would be showing that much more verbose error message every time a customer of RV-Match makes one of these easy-to-make mistakes. Again, we could probably work around this with sufficient engineering, but it starts to become really complex, which is less than ideal from a design perspective. I would be much more inclined to simply include something in a troubleshooting section of the README talking about these two errors and explaining what you have described there, rather than changing the error message in the tool itself.

Does that seem okay to you?

grosu:
OK, sounds like a good plan.

dwightguth:
Okay I updated the README as well, please take a look.

I think maybe while we are on this issue we should also think about updating the license like we talked about. The sooner change the license for the C semantics, the safer we will be against the possibility that someone will decide to reuse our work without us seeing money.

I did a lot of searching on the web and the closest I could find to a standard license that does pretty much what we want is this: http://pigale.sourceforge.net/license_Qt.html

Can you please take a stab at adapting this license to suit our needs? Essentially what we want is that users can distribute and use adaptations of the C semantics as long as they do not profit from their use and they attribute us as the original authors. We also probably want to include a license exception for people who have purchased an RV-Match license from Runtime Verification.

ellisonch:
You probably know this, but because of the license you have, people can
always take the current or older versions and use them in their own
commercial products.  If you decide to relicense, it will only effect
versions going forward from that point.

*As far as I understand, and I'm obviously not a lawyer.

-Chucky

ellisonch:
You may also want to consider continuing to release the base project under
the current license, and make an additional plugin, or module, or whatever
that represents the business component.  You could even keep the source of
the business component closed, or prohibit commercial uses.  That way, your
bug fixes and improvements to the base project can still be released truly
openly, but your secret sauce stays protected.

Food for thought.

-Chucky

dwightguth:
@grosu please review the adapted license that I wrote.

chathhorn:
This looks fine to me, although note that the old license was kept because the wording of it seems to require that it be kept.

dwightguth:
Hmm. @ellisonch do you want to try to enforce copyright on your portions of the semantics? We can probably find a way to include the old license in a way that makes it clear that the current version is a derivative work based on the old version that you have copyright on, if it's something that matters to you.

ellisonch:
I have copyright for the original semantics, and used the MIT license.  My
understanding of what happened is that you guys wanted to switch to using
the UIUC license, which you can do for the additions you made, but
according to the terms of the MIT license, you just need to keep the
original copyright message and license around (the same way you do for CIL;
in fact, I think that's exactly how you should do it; something like,
portions copyright blah blah under license blah blah).

It's not a huge deal, though it is required by the terms of the license.

More importantly, I do hope you can find a way to keep bug fixes and
semantics improvements using an open license like UIUC or MIT, even if you
protect your secret sauce with a non-commercial license.  Maybe academic
work (work not being paid for by the company) can continue to be released
in the project using the UIUC license, whereas people doing RV work gets
put in a separate module or plugin that is released under a commercial-free
license.  Obviously, this is only a hope, as you guys can do whatever you
want, but I'd like to see academic work continue to be released under
permissive licenses.

-Chucky

dwightguth:
@ellisonch does this look good to you? I re-added the old license but added a reference to why it is necessary by saying that portions of the semantics are your copyright and referring to the license file.

As far as your concern, the reason we are switching licenses is that we would like the C semantics to eventually grow to the point where KCC in its original form is capable of checking undefinedness in any program as long as you can use  KCC to build a version of the standard library (e.g. gcc). If we do this, but maintain a permissive license, then we no longer have any competitive advantage whatsoever and any secret sauce we may have developed essentially becomes irrelevant and can be bypassed. So in order to do this, we would find it necessary to make portions of the semantics itself be proprietary and stored under a different build system, which really makes things quite complex. It would also have the consequence that the version of the tool for academic use would not nearly be as good as the version we sell, which is actually a worse state of affairs for academic purposes than we would be able to provide by splitting up the semantics.

One other thing to note is that RV actually possesses an exclusive license from UIUC to commercialize software for which UIUC owns the copyright which was developed by the FSL group. So technically speaking we have actually already had license incompatibility and the way to deal with that is to make the C semantics license be noncommercial.

dwightguth:
Anyone have any more suggestions to make before I merge this? @ellisonch I know it's not really what you want but are you satisfied that we're allowed to do what we've done? @grosu any more comments? @chathhorn do you want to take another look now that I've rearranged things some?

chathhorn:
Looks good to me.

README.md
```diff
@@ -24,58 +24,51 @@ understand this project:
...
-torture test suite, use the following command from the [tests][] directory:
-```
-$ make torture
-```
-And to run the undefinedness test suite:
-```
-$ make undef
-```
+We recommend if you wish to execute such programs that you contact Runtime Verification,
+Inc, which sells a tool RV-Match based on this semantics which is capable of executing
```
grosu: sells => licenses

ok with me to merge after that.

dwightguth:
@ellisonch if you don't have more comments today I am probably going to merge this this afternoon.

PR #131, opened 2015-10-19
Fixes To Error Codes

msaxena2:
Fixes to Error Codes. Duplicate error codes with different error messages fixed. WIP.

PR #147, opened 2015-10-27
New errors2

dwightguth:
@chathhorn please review

It's grouped by commit so you should be able to review individual commits in the diff without it blowing up on you for being too big.


dwightguth:
lgtm. This is @msaxena2's code originally. See #131 for original history.

chathhorn:
This looks great!

PR #154, opened 2015-11-02
Fixes2

dwightguth:
@chathhorn @grosu please review.

@grosu note that there is still inconsistency between " " and "="  but it is now based completely on the same behavior that GCC has.


chathhorn:
Looks good to me!

grosu:
lgtm

PR #155, opened 2015-11-02
update kcc to use -profile and -v for switching

dwightguth:
@chathhorn @grosu please review


scripts/kcc
```diff
@@ -26,14 +26,24 @@ my @srcFiles = ();
 my @tempFiles = ();
 
 my $distDirectory = dirname(rel2abs($0));
+our $profile;
```
chathhorn: I notice some others have crept in as well, but shouldn't all the variables in this file be declared using "my" so they're lexically scoped to this file?

chathhorn:
Looks fine to me otherwise, assuming the tests pass.

scripts/kcc
dwightguth: @chathhorn this is done so that they can be accessed from the actions on the Getopt::Declare.

dwightguth:
The build is failing because of a bug with the makefile that should be fixed by the installer pull request. All tests are actually passing, so I am going to merge this.

PR #162, opened 2015-11-06
remove noncommercial from license

dwightguth:
@grosu please review

@ellisonch FYI. We decided to go a different route with the licensing after all. There is definitely some potential for lost revenue down this new path, but if nothing else we should be able to make money for the foreseeable future with the current proprietary components we have, and we have discussed plans to make a few less-critical components also proprietary.


grosu:
lgtm

PR #166, opened 2015-11-10
Refactor handling of local decls.

### LINKED ISSUE #139: Redeclaring the same identifier within a block not resulting in an error
This following program,
``` C
int main() {
  int a;
  int a;
}
```
produces no error with kcc. Also executes without error. 

Shouldn't an error like, "redeclaration of identifier with no linkage" be thrown here?

chathhorn:
Fixes #139. @dwightguth please review.

### DIFF
--- a/examples/simple/violates-constraint/CV-FD1-bad-static.output
+++ b/examples/simple/violates-constraint/CV-FD1-bad-static.output
@@ -1,7 +1,7 @@
 File: CV-FD1-bad-static.c
-Line: 1
-Error: CV-FD1
-Description: duplicate parameters in function definition.
+Line: 3
+Error: CV-TDL6
+Description: redeclaration of x with no linkage.
 Type: Constraint violation.
 See also: C11 sec. 6.7:3
 Translation failed. Run kcc -d CV-FD1-bad-static.c to see commands run.
--- a/semantics/language/translation/decl/local.k
+++ b/semantics/language/translation/decl/local.k
@@ -29,14 +29,13 @@ module C-DECL-LOCAL
           <types> M:Map => M[X <- T] </types>
           <external-types>... X |-> T:Type ...</external-types>
           <external-defs>... X |-> Base:SymBase ...</external-defs>
-          <block-history> ListItem(BlockNum:Int) ...</block-history>
           <curr-tu> TU:String </curr-tu>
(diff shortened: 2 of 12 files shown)


semantics/language/translation/decl/local.k
```diff
@@ -179,27 +169,25 @@ module C-DECL-LOCAL
           ~> defineAndInit(_, _, initValue(_, T:Type, _))
           requires notBool isCompleteType(T)
           [structural]
+     rule <k> (.K => CV("TDL6","redeclaration of " +String idToString(X)
```
dwightguth: I think this should have the same error code as TDL4 since it has the same error message and refers to the same paragraph in the standard.

tests/unit-fail-compilation/dup_params1.ref
```diff
@@ -1,7 +1,7 @@
 File: dup_params1.c
 Line: 1
-Error: CV-FD1
-Description: duplicate parameters in function definition.
+Error: CV-TDL6
```
dwightguth: I believe you will also need to update CV-FD1 in the examples/simple directory. Do you know if it's even possible to trigger this error anymore or has it been completely superceded?

chathhorn: It still gets triggered in the case of old style defs with duplicate ids in
the id list (but not duplicate declarations). So many caveats...

dwightguth: Okay. Then yeah, make sure you check the example to ensure that it is still triggering the right error and update it if necessary.

chathhorn:
I'm going to go ahead and merge this, more decl refactorings forthcoming.

examples/simple/violates-constraint/CV-FD1-bad-static.output
```diff
@@ -1,7 +1,7 @@
 File: CV-FD1-bad-static.c
-Line: 1
-Error: CV-FD1
-Description: duplicate parameters in function definition.
+Line: 3
+Error: CV-TDL6
```
dwightguth: This isn't what you're supposed to do. The point of this file is it's a file that throws the FD1 error. If this example no longer throws an error, you need to construct a new example that does. And if no such example can be constructed, then you need to delete the example from the suite.

PR #169, opened 2015-11-11
Updated examples.

chathhorn:
This might get a bit painful to keep consistent... @dwightguth please review.


dwightguth:
lgtm. Eventually we want to have test coverage for this I just haven't gotten around to it yet.

PR #173, opened 2015-11-16
fix confluence issue with storage class specifiers on parameters

dwightguth:
@chathhorn please review. It's a little ambiguous whether this is a correct change because the constraint violation says that a parameter declaration shall not have a storage class specifier other than register and it's not 100% clear whether `(void)` counts as a parameter declaration in that sense. If yes, then this change is technically incorrect. But I think since we report an error either way it should probably be fine, and I like the advantage of having confluent rules for this case because it fixes the case where tests can fail based on the order of rules in the compiled definition.


chathhorn:
Looks fine if the tests pass.

PR #175, opened 2015-11-17
Fix issue with VLAs in sizeof.

chathhorn:
@dwightguth please review.


semantics/language/translation/typing/expr.k
```diff
@@ -199,10 +200,19 @@ module C-TYPING-EXPR
...
           requires areCompatible(stripQualifiers(T), stripQualifiers(T'))
+               andBool notBool (isVariablyModifiedType(T) orBool isVariablyModifiedType(T'))
      rule typeof(_ ? t(Mods1:Set, pointerType(t(Mods1':Set, S:SimpleType))) : t(Mods2:Set, pointerType(t(Mods2':Set, void))) => t(Mods1 Mods2, pointerType(t(Mods1' Mods2', void))))
           requires notBool isFunctionType(t(Mods1', S))
+               andBool notBool isVariablyModifiedType(t(Mods1', S))
      rule typeof(_ ? t(Mods1:Set, pointerType(t(Mods1':Set, void))) : t(Mods2:Set, pointerType(t(Mods2':Set, S:SimpleType))) => t(Mods1 Mods2, pointerType(t(Mods1' Mods2', void))))
           requires notBool isFunctionType(t(Mods2', S))
+               andBool notBool isVariablyModifiedType(t(Mods2', S))
+
+     rule (.K => UNDEF("TTE1",
```
dwightguth: is this the only case in this class where a composite type is taken?

chathhorn: It's the only place in that file.

But it looks like technically (_ ? NULL : ((*) [x]) NULL) probably doesn't require taking the composite type according to 6.5.15:6 (and should just have the type of (*)[x]), although that's pretty subtle...

dwightguth: Yes, that sounds right, because NULL is a null pointer constant. So the example you constructed so far should actually be correct with the type as you describe. However, if I were to do something like:
```
void foo(int (*x)[*], int y) {
  sizeof(0 ? x : (int (*) [y]) NULL);
}
```

Then it would take the composite type and be undefined.

chathhorn: Yeah, I agree, I'm reworking it.

dwightguth:
Is this ready for re-review yet?

chathhorn:
It should be good now, but you might wait for the tests to finish.

chathhorn:
Closing for cleanup/rebase.

PR #177, opened 2015-11-20
more fixes for benchmark

dwightguth:
@chathhorn please review


semantics/language/common/expr/reference.k
```diff
@@ -67,6 +67,7 @@ module C-COMMON-EXPR-REFERENCE
      }*/
      rule *(tv(Loc:SymLoc, T:PointerType))
           => checkRestrict(Loc)
+          ~> checkLoc(Loc)
```
chathhorn: Sure this is necessary? Bad addresses should in theory be caught well before this point.

dwightguth: It seems to happen only if the address is of static duration. Do you know why that might be?

chathhorn:
This looks good.

semantics/language/common/expr/reference.k
chathhorn: Yeah, I suppose ids for static duration stuff will be replaced by their addresses during translation. It might be better to tone that down a bit.

PR #178, opened 2015-11-23
Improve treatment of VLAs in sizeof & params.

chathhorn:
@dwightguth please review.


examples/simple/violates-constraint/CV-EFNC3-bad.output
```diff
@@ -1,9 +0,0 @@
-Error: CV-EFNC3
```
dwightguth: can you please confirm we still catch this error under a different error code?

semantics/language/translation/error.k
```diff
@@ -19,57 +20,37 @@ module C-TRANSLATION-ERROR
                +String "\nLine: " +String Int2String(Line)
                +String "\n")
           ~> HALT ...</k>
+          <curr-scope> S:Scope </curr-scope>
           <curr-program-loc>
                CabsLoc(File:String, Line:Int, _, _)
           </curr-program-loc>
           <result-value> _ => 1 </result-value>
-          requires notBool isCSV
+          requires notBool isCSV andBool S =/=K noEvalScope
```
dwightguth: what exactly happens here if S ==K noEvalScope? Does it get stuck or what?

semantics/language/translation/typing/expr.k
```diff
@@ -189,20 +187,130 @@ module C-TYPING-EXPR
...
+     rule (.K => elaborate(scope(noEvalScope, elabResult(E1)), scope(noEvalScope, elabResult(E2))))
+          ~> typeof(_ ? E1:KItem : E2:KItem)
+
+     syntax KItem ::= elabResult(K)
+     context elabResult(HOLE:KItem => reval(HOLE)) [result(RValue)]
+
+     rule <k> elabResult(V:RValue) => .K ...</k>
+          <elab> .K => V ...</elab>
+
+     rule <k> EXIT(_) => recover ...</k>
```
dwightguth: oh I see here, never mind.

semantics/language/translation/env.k
```diff
@@ -105,4 +106,32 @@ module C-ENV
                ListItem(L:LocalCell) => .List
           ...</block-stack>
 
+     rule recover ~> (K:KItem => .K)
+          requires K =/=K popLocals
+     rule (recover => .K) ~> popLocals
```
dwightguth: one thing I'm not entirely clear on here though is, if the conditional expression contains an error, how does "recover" eventually lead it back to having a type?

dwightguth:
I don't see anything obviously wrong but there's still a few minor things I'm unclear on with respect to how this diff works. I see you also did some refactoring of function definitions and the unspecified variable length array type. Is there anything else in this diff I should focus on as a change you made?

examples/simple/violates-constraint/CV-EFNC3-bad.output
chathhorn: This one is UB-CB1 now.

semantics/language/translation/env.k
chathhorn: It just comes back as ".K" if there's an error while calculating the type.

chathhorn: Rather, if there's an error while elaborating.

chathhorn:
I can't think of anything in particular. I think this makes our treatment of VLAs in params actually totally correct AFAICT.

PR #186, opened 2015-12-04
create profile for rv-match to use so that default profile can have c…

dwightguth:
…ore extensions

@chathhorn please review


chathhorn:
Looks good.

PR #194, opened 2015-12-17
ignore windows versus linux line breaks in test outputs

dwightguth:
@pdaian please review

This is the first of two diffs that hopefully should fix the windows build.


pdaian:
lgtm

PR #197, opened 2015-12-24
New Examples for Posix Threads + Memory Lint Errors

msaxena2:
WIP. Not yet ready.

### DIFF
--- a/examples/simple/Error_Codes.csv
+++ b/examples/simple/Error_Codes.csv
@@ -1,171 +1,171 @@
-,,,
-,The C11 semantic errors follow the template: Error_Type-Error_Code.,,
-,"The Error_Type can be one of: UB (Undefined Behavior), USP (Unspecified Behavior), CV (Constraint Violation), or IMPL (Implementation Specific Behavior)",,
-,The Error_Code is a unique code used to identify a particular error.,,
-,,,
-Error,Description ,References,Type
-UB-CB1,Types of function call arguments aren't compatible with declared types after promotions.,"6.5.2.2:6, J.2:1 item 39",Undefined Behavior
-UB-CB2,Function call has fewer arguments than parameters in function definition.,"6.5.2.2:6, J.2:1 item 38",Undefined Behavior
-UB-CB3,Function call has more arguments than parameters in function definition.,"6.5.2.2:6, J.2:1 item 38",Undefined Behavior
-UB-CB4,Function defined with no parameters called with arguments.,"6.5.2.2:6, J.2:1 item 38",Undefined Behavior
-UB-CCV1,Signed integer overflow.,"6.5:5, J.2:1 item 36",Undefined Behavior
-UB-CCV3,Conversion to integer from float outside the range that can be represented.,"6.3.1.4:1, J.2:1 item 17",Undefined Behavior
-UB-CCV4,Floating-point value outside the range of values that can be represented after conversion.,"6.3.1.5:1, J.2:1 item 18",Undefined Behavior
-UB-CCV5,Casting empty value to type other than void.,"6.3.2.2:1, J.2:1 item 23",Undefined Behavior
-UB-CCV6,Casting void type to non-void type.,"6.3.2.2:1, J.2:1 item 23",Undefined Behavior
-UB-CCV7,Conversion from pointer to integer of a value possibly unrepresentable in the integer type.,"6.3.2.3:6, J.2:1 item 24",Undefined Behavior
-UB-CCV11,Conversion to a pointer type with a stricter alignment requirement (possibly undefined).,"6.3.2.3:7, J.2:1 item 25",Undefined Behavior
-UB-CCV12,Floating-point overflow.,"6.5:5, J.2:1 item 36",Undefined Behavior
-UB-CEA1,A pointer (or array subscript) outside the bounds of an object.,"6.5.6:8, J.2:1 item 46",Undefined Behavior
-UB-CEA5,Computing pointer difference between two different objects.,"6.5.6:9, J.2:1 item 48",Undefined Behavior
-UB-CEA2,Pointer difference outside the range that can be represented by object of type ptrdiff_t.,"6.5.6:9, J.2:1 item 50",Undefined Behavior
-UB-CEA6,"Unspecified value in signed addition, possible overflow.","6.5:5, J.2:1 item 36",Undefined Behavior
-UB-CEB2,The right operand in a bitwise shift is negative.,"6.5.7:3, J.2:1 item 51",Undefined Behavior
-UB-CEB3,The right operand in a bitwise shift is greater than or equal to the bit width of the left operand.,"6.5.7:3, J.2:1 item 51",Undefined Behavior
-UB-CEB4,The left operand in a bitwise left-shift is negative.,"6.5.7:4, J.2:1 item 52",Undefined Behavior
-UB-CEB6,The right operand in a bitwise shift is negative.,"6.5.7:3, J.2:1 item 51",Undefined Behavior
-UB-CEB12,Potential negative zero produced via bitwise operations (this could be a false positive).,"6.2.6.2:4, J.2:1 item 14",Undefined Behavior
-UB-CEB7,The right operand in a bitwise shift is greater than or equal to the bit width of the left operand.,"6.5.7:3, J.2:1 item 51",Undefined Behavior
-UB-CEE2,Indeterminate value used in an expression.,"6.2.4, 6.7.9, 6.8, J.2:1 item 11",Undefined Behavior
-UB-CEE1,Address of an array declared with register storage class used in an expression.,"6.3.2.1:3, J.2:1 item 22",Undefined Behavior
-UB-CEE4,Referring to an object outside of its lifetime.,"6.2.4:2, J.2:1 item 9",Undefined Behavior
-UB-CEE3,Found pointer that refers outside the bounds of an object + 1.,"6.3.2.1:1, J.2:1 item 19",Undefined Behavior
-UB-CEM1,Attempting to access a member in an atomic struct.,"6.5.2.3:5, J.2:1 item 42",Undefined Behavior
-UB-CEM2,Attempting to access a member in an atomic union.,"6.5.2.3:5, J.2:1 item 42",Undefined Behavior
-UB-CEMX1,Division by 0.,"6.5.5:5, J.2:1 item 45",Undefined Behavior
-UB-CEMX2,Modulus by 0.,"6.5.5:5, J.2:1 item 45",Undefined Behavior
-UB-CEMX3,Signed integer overflow.,"6.5:5, J.2:1 item 36",Undefined Behavior
-UB-CER3,Dereferencing a null pointer.,"6.5.3.2:4, J.2:1 item 43",Undefined Behavior
-UB-CER4,Dereferencing a pointer past the end of an array.,"6.5.6:8, J.2:1 item 47",Undefined Behavior
-UB-CERL1,Cannot compare pointers with different base objects using '<'.,"6.5.8:5, J.2:1 item 53",Undefined Behavior
-UB-CERL2,Cannot compare pointers with different base objects using '<='.,"6.5.8:5, J.2:1 item 53",Undefined Behavior
-UB-CERL3,Cannot compare pointers with different base objects using '>'.,"6.5.8:5, J.2:1 item 53",Undefined Behavior
-UB-CERL4,Cannot compare pointers with different base objects using '>='.,"6.5.8:5, J.2:1 item 53",Undefined Behavior
-UB-CMW1,Trying to modify a string literal.,"6.4.5:7, J.2:1 item 33",Undefined Behavior
-UB-CTM1,Variable-length array with length that evaluates to non-positive value at execution.,"6.7.6.2:5, J.2:1 item 75",Undefined Behavior
-UB-ECL3,An object which has been modified is accessed through an expression based on a restrict-qualified pointer and another lvalue not also based on said pointer.,"6.7.3.1:4, J.2:1 item 68",Undefined Behavior
-UB-ECL4,An object which has been modified is accessed through an expression based on a restrict-qualified pointer to a const-qualified type.,"6.7.3.1:4, J.2:1 item 68",Undefined Behavior
-UB-CEER1,Trying to read through a null pointer.,"6.3.2.1:1, J.2:1 item 19",Undefined Behavior
-UB-CEER4,Have a read-write datarace.,"5.1.2.4:25, J.2:1 item 5",Undefined Behavior
-UB-CEER5,Have a write-write datarace.,"5.1.2.4:25, J.2:1 item 5",Undefined Behavior
-UB-CID2,No definition found for identifier .,"6.3.2.1:1, J.2:1 item 19",Undefined Behavior
-UB-EEA1,Assignment from non-exactly-overlapping overlapping read.,"6.5.16.1:3, J.2:1 item 54",Undefined Behavior
-UB-EEA2,Attempt to alias restrict qualified pointers associated with the same block.,"6.7.3.1:4, J.2:1 item 69",Undefined Behavior
-UB-EEA3,Attempt to alias restrict qualified pointers associated with incompatible blocks.,"6.7.3.1:4, J.2:1 item 69",Undefined Behavior
-UB-EEA4,Passing a pointer not to an array object as a function parameter with a static-qualified array size.,"6.7.6.3:7, J.2:1 item 77",Undefined Behavior
-UB-EEA5,Passing a pointer to an array object of fewer elements than required by a function parameter with a static-qualified array size.,"6.7.6.3:7, J.2:1 item 77",Undefined Behavior
-UB-EEC1,Incompatible variably modified types in a conditional expression.,6.7.6.1:6,Undefined Behavior
-UB-EFNC1,Attempt to call the function through a pointer with incompatible type.,"6.5.2.2:9, J.2:1 item 41",Undefined Behavior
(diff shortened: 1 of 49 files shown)


msaxena2:
@dwightguth @grosu Please review.

examples/simple/Error_Codes.csv
```diff
@@ -95,6 +95,14 @@ UB-STDLIB1,"Called realloc on memory not allocated by malloc, or already freed."
 UB-STRING1,Null pointer passed to strcpy.,"7.1.4:1, J.2:1 item 108",Undefined Behavior
 UB-STDLIB2,"Called free on memory not allocated by malloc, or already freed.","7.22.3.3:2, J.2:1 item 179",Undefined Behavior
 UB-TTE1,program requires the formation of a composite type from a variable length array type whose size is specified by an expression that is not evaluated.,"6.2.7:3, J.2:1 item 16",Undefined Behavior
+UB-PTHREAD1,Initializing an already initialiized mutex,POSIX 2008 sec. pthread_mutex_init,Undefined Behavior
```
dwightguth: I'd like to see an extra column with like an X or something in it for those errors that are only caught with RV-Match.

examples/simple/lint/L-EXC3-bad.c
```diff
@@ -0,0 +1,20 @@
...
+
+void func(int number)
+{
+    printf("func called with number: %i\n", number);
+}
+
+int main()
+{
+    void (*func_ptr)(int) = &func;
+    if(func_ptr != ((void *) NULL))
```
dwightguth: I think just NULL should be fine because NULL is usually already of type void *

dwightguth: also I believe the idea behind this error is it occurs with functions that take no arguments when the user has forgotten to add the () that causes the function to actually be invoked. So I think the good example here should be that we are actually calling func().

examples/simple/posix/UB-PTHREAD5-bad.c
```diff
@@ -0,0 +1,17 @@
...
+#include<stdlib.h>
+
+
+pthread_mutex_t * lock_ptr;
+
+
+int main(void)
+{
+    //Calling pthread_mutex_lock without a call to pthread_mutex_init
+    lock_ptr = malloc(sizeof(pthread_mutex_t));
```
dwightguth: I think to make this more interesting what we should do is to initialize the mutex with pthread_mutex_init, but then copy it with memcpy and call the method on the copied mutex.

dwightguth:
looks pretty good. there's a couple things I would do to make this a little more interesting, but otherwise it seems solid to me.

msaxena2:
@dwightguth Please re-review.

dwightguth:
looks good. If the tests pass and @grosu is okay with it go ahead and merge this.

examples/simple/lint/L-CEIE1-good.c
```diff
@@ -0,0 +1,12 @@
...
+
+#include <stdlib.h>
+#include <stdio.h>
+
+int main()
+{
+    int * number_ptr = malloc(sizeof(int));
+    *number_ptr = 10;
+    free(number_ptr);
+    printf("No memory Leaks, Execution Successful\n");
```
grosu: Be uniform: either capitalize all words or none.  In fact, since the bad example was minimal, you can also make this one minimal (ie, remove assignment and printf).  Or did you have a reason for which you added the additional statements?

examples/simple/lint/L-EXC3-bad.c
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2015 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+
+#include <stdlib.h>
+#include <stdio.h>
+
+void * func()
+{
+    int * retval = malloc(sizeof(int));
+    *retval = 1;
+    return retval;
```
grosu: Don't we have a L-CEIE1 error/warning here, that memory is malloced but not freed?

dwightguth: No because this function is never actually called.

grosu: I would either remove the malloc statement, or otherwise add a comment in the code saying that the function is not called, so no memory leaks, both in the bad and in the good example. That's because some static analysis tools may report an error here, and we do not want them to think that we "missed" the error.  Something as simple as:

// No memory leak reported, because the function is never invoked.

dwightguth: yeah I'd be inclined to just make the function return a pointer to a statically allocated object, eg:
```
int x;
void *func() { return &x; }
```

dwightguth: and then do that for both bad and good

examples/simple/lint/L-STDLIBE2-good.c
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2015 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+
+#include <stdlib.h>
+#include <stdio.h>
+
+int main()
+{
+    size_t array_size = 20;
```
grosu: Does this need to be size_t instead of int?  One would expect the -good programs to only fix the -bad ones minimally.

dwightguth: good catch

examples/simple/lint/L-STDLIBE3-good.c
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2015 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+
+#include <stdlib.h>
+#include <stdio.h>
+
+int main()
+{
+    size_t array_size = 20;
```
grosu: same here

examples/simple/posix/UB-PTHREAD2-bad.output
```diff
@@ -0,0 +1,12 @@
+Error: UB-PTHREAD2
+Description: Destroying a locked mutex.
+Locked here:
+  at pthread_mutex_lock(UB-PTHREAD2-bad.c:15)
+  at locker_thread(UB-PTHREAD2-bad.c:15)
+  at <file-scope>(<unknown>)
+
```
grosu: Is this white line necessary?  I initially thought that there are two error messages.  Others may be confused, too.

examples/simple/posix/UB-PTHREAD4-bad.output
```diff
@@ -0,0 +1,12 @@
+Error: UB-PTHREAD4
+Description: Attempting to lock a PTHREAD_MUTEX_DEFAULT mutex that is already locked by the current thread.
+Locked here:
+  at pthread_mutex_lock(UB-PTHREAD4-bad.c:15)
+  at locker_thread(UB-PTHREAD4-bad.c:15)
+  at <file-scope>(<unknown>)
+
```
grosu: same here: is the whiteline necessary, or follows some more general convention?  if not, i would remove it.

examples/simple/posix/UB-PTHREAD5-good.c
```diff
@@ -0,0 +1,19 @@
...
+#include <pthread.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+pthread_mutex_t lock;
+
+int main(void)
+{
+
```
grosu: any reason for this whiteline?  you typically didn;t have whitelines at the beginning of functions.  also, not clear to me when you put one whiteline and when you put more.  no big deal, but in general it is a good idea to follow consistent formatting and rules, because that shows seriousness and professionalism.

examples/simple/posix/UB-PTHREAD2-bad.output
dwightguth: @msaxena2 go ahead and remove this line and I will fix the code in the upcoming pull request I am working on that deals with examples.

examples/simple/posix/UB-PTHREAD6-bad.output
```diff
@@ -0,0 +1,7 @@
+Error: UB-PTHREAD6
+Description: Attempting to unlock a PTHREAD_MUTEX_DEFAULT or PTHREAD_MUTEX_NORMAL mutex which is not robust and which has not been locked by the current thread.
```
grosu: This description is a bit hard to follow.  Can you remove `which is not robust and`?  Since it is an undefined behavior, it is sort of understood that it is not robust.  Or am I misunderstanding it?

examples/simple/posix/UB-PTHREAD6-bad.c
```diff
@@ -0,0 +1,44 @@
...
+#include<stdlib.h>
+
+
+
+pthread_t thread_id;
+pthread_mutex_t lock;
+
+void * locker_thread(void * ptr) 
+{
+    //unlocking an unlocked thread
```
grosu: I am all for such comments, but we should aim at consistency: either we add such comments to all -bad examples, or to none, and we should either start all of them with a capital letter or none, and we should have a period at the end everywhere or nowhere.  actually I'm not very sure about the first, it kind of makes sense to add such comments selectively, where things are tricky.

examples/simple/posix/UB-PTHREAD8-good.c
```diff
@@ -0,0 +1,47 @@
...
+
+   
+    
+    pthread_join(thread_id, NULL);
+    pthread_mutex_destroy(&lock);
+    
+    printf("Successful Execution\n");
+    
+    return 0;
+}
```
grosu: The -good one seems a lot larger and different than the -bad one.  Can they get closer to each other?  I think people will like to see quickly the fix.

examples/simple/posix/UB-PTHREAD6-bad.output
dwightguth: @grosu  a robust mutex is a special type of mutex in POSIX 2008. Attempting to unlock a robust mutex does not result in undefined behavior, hence the error message. If you can think of another way of rephrasing this, I am open to suggestions.

grosu:
I have no other comments.  Very nice examples!  In fact, I think having people go through our canonical examples is a good idea, because they both learn C better and appreciate RV-Match more.  I'm going to add a sentence or two about that in the documentation.  Thank you, guys, very good work!

examples/simple/posix/UB-PTHREAD6-bad.output
grosu: But the type of the error is Undefined behavior.  If this was a warning and the type potentially undesired behavior, then I would have raised no objections.

dwightguth: It is considered undefined behavior, yes. Specifically it is considered undefined by the POSIX 2008 standard to unlock a mutex you do not own and have locked, unless the mutex is a  robust mutex, in which case it is well defined and returns a specific error code.

grosu: OK then.  What I didn't know was that `robust` is a technical term in this context.  Hopefully others will not get confused.

msaxena2:
@dwightguth @grosu Made some more changes. Please re-review.

examples/simple/lint/L-CCVE2-bad.output
```diff
@@ -0,0 +1,5 @@
+Warning: L-CCVE2
+Description: conversion to unsigned integer outside the range that can be represented.
```
grosu: The description is still not capitalized.

dwightguth:
looks good. seems like you addressed all the comments. You tested all the examples and they seem to work as intended? If so, and if @grosu is okay, I am fine with this being merged.

examples/simple/lint/L-CEIE1-bad.c
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2015 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+
+#include <stdlib.h>
+#include <stdio.h>
+
+int main()
```
grosu: You had two lines before the main function until now, and now you have one line.

@msaxena2 , can you please go through my previous comments and address them one by one, telling whether you took care of it and how, or whether you have not and why?  Otherwise I'm wasting a lot of time re-checking everything that I've already reported.

examples/simple/lint/L-CCVE1-bad.output
```diff
@@ -0,0 +1,5 @@
+Warning: L-CCVE1
+Description: unsigned integer overflow.
```
dwightguth: this and the one @grosu mention should be capitalized. I will make a PR to handle the code change.

examples/simple/lint/L-EXC3-bad.c
```diff
@@ -0,0 +1,21 @@
...
+    return &retval; 
+}
+
+int main()
+{
+    void * (*func_ptr)() = &func;
+    if(func_ptr != NULL)
+    {
+       printf("Hello, World!\n"); 
+
```
grosu: why a white line here?

examples/simple/lint/L-PTHREAD9-bad.output
```diff
@@ -0,0 +1,8 @@
+Warning: L-PTHREAD9
+Description: Terminating a thread while holding lock on mutexes. Locked here:
+  at pthread_mutex_lock(L-PTHREAD9-bad.c:15)
+  at locker_thread(L-PTHREAD9-bad.c:15)
+  at <file-scope>(<unknown>)
+
```
grosu: why whiteline?

examples/simple/lint/L-PTHREAD9-good.c
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) 2015 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+
+#include<stdio.h>
+#include<string.h>
+#include<pthread.h>
+#include<stdlib.h>
+
+
```
grosu: why 3 whitelines?

examples/simple/lint/L-STDLIBE2-good.c
grosu: so, do we change this to int or not?

examples/simple/lint/L-STDLIBE2-good.c
```diff
@@ -0,0 +1,15 @@
...
+    int * small_int_array = malloc(array_size * sizeof(int));
+    if(small_int_array != NULL)
+    {
+
```
grosu: why whiteline?

examples/simple/lint/L-STDLIBE3-bad.c
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2015 Runtime Verification, Inc. (RV-Match team). All Rights Reserved.
+
+#include <stdlib.h>
+#include <stdio.h>
+
+int main()
+{
+    size_t array_size = 1024 * 1024 * 1024;
```
grosu: int?

examples/simple/posix/UB-PTHREAD1-bad.c
```diff
@@ -0,0 +1,24 @@
...
+
+#include <stdio.h>
+#include <pthread.h>
+#include <stdlib.h>
+
+pthread_mutex_t lock;
+
+int main(void)
+{
+
```
grosu: why whiteline?

examples/simple/posix/UB-PTHREAD6-bad.c
```diff
@@ -0,0 +1,43 @@
...
+int main(void)
+{
+    int err = 0;
+    if (pthread_mutex_init(&lock, NULL) != 0)
+    {
+        printf("\n mutex init failed\n");
+        return 1;
+    }
+
+
```
grosu: 3 whitelines?

examples/simple/lint/L-CEIE1-good.c
msaxena2: I agree the printf was unnecessary; it's been removed.

examples/simple/lint/L-EXC3-bad.c
msaxena2: Malloc has been replaced by a statically allocated object in both -good and -bad versions.

examples/simple/lint/L-STDLIBE2-good.c
msaxena2: for now, I've kept size_t in both good and bad versions. We can change it if needed.

examples/simple/lint/L-CEIE1-bad.c
msaxena2: I've added another whitespace here to adhere to a consistent format, described in one of the later comments.

examples/simple/lint/L-PTHREAD9-good.c
msaxena2: Changed to 2 whitespaces.

examples/simple/lint/L-STDLIBE2-good.c
msaxena2: Removed the whitespace here.

examples/simple/posix/UB-PTHREAD1-bad.c
msaxena2: Addressed by removing the whiteline.

examples/simple/posix/UB-PTHREAD6-bad.c
msaxena2: Addressed by removing the whiteline.

msaxena2:
@grosu @dwightguth Please re-review. Changes made since previous review:
* I've addressed all the whiteline related comments that had been made in the previous review. The files now have 2 whitelines between the last '#include' and the following line of code, and a single whiteline between different function definitions. Within function blocks, I've removed random occurrences of whitelines. I've done this for all files to ensure uniformity.
* Both good and bad version of STDLIBE2, STDLIBE3 now use size_t instead of int. I'm open to reverting them back to int.

grosu:
lgtm

PR #200, opened 2015-12-29
Javabackend

dwightguth:
@chathhorn @msaxena2 please review

This includes both my cleaned-up version of Manasvi's changes to the code to work with the java backend, and a number of small fixes designed to make error recovery recover in more cases.

Should be the first of a two part diff, the other part being in the rv match repo. They should probably be merged together.


msaxena2:
Lgtm.

chathhorn:
Looks good!

PR #202, opened 2015-12-29
Presentation improvements

grosu:
I've made some changes to the C semantics, that go in synch with changes I've made to the RV-Match docs (will make that PR shortly after this one).

@dwightguth please revise; feel free to fix English however you find fit.


dwightguth:
This diff is too big for me to review in its current form. I need you to move all of the rename-without-modifications into one commit and all of everything else into a different commit. You can do this by refactoring the history and doing a force push. Let me know if you need help.

The other issue here is that we have several open pull requests that also modify examples, which will not merge cleanly with this diff. so we will not be able to merge this diff until those other pull requests are merged and then this diff is manually updated to make sure the renames are still going through cleanly.

examples/demo/6-int-overflow.c
```diff
@@ -1,9 +1,21 @@
 // KCC is also able to detect integer overflow operations that occur while the
-// program is executing.
+// program is executing.  According to the C standard, integer overflow
+// operations fall under the category of implementation defined behavior,
```
dwightguth: this is not quite correct. Try the following paragraph instead:

KCC is also able to detect integer overflow operations and assignments of values to data types triggering loss of precision that occur while the program is executing. According to the C standard, integer overflow (an arithmetic operation whose result does not fit in the specified type) is considered undefined behavior. However, in the following program, a short int is automatically promoted to the int type at every multiplication, and therefore what occurs is not overflow but merely a conversion to a type that cannot store the specified value. The C standard considers this case to be implementation defined behavior, which is unspecified behavior where each implementation documents how the choice is made. KCC can be instantiated with different profiles corresponding to different implementation choices (type kcc -v to see the existing profiles; contact https://runtimeverification.com/support if you are interested in a specific profile that is not available). For example, while with conventional compilers the program below reports no errors when executed, with the default kcc profile (x86_64 on linux) it reports an error, but no error is reported if you declare variable a as an int instead of a short int, or if you iterate 14 times instead of 15. If you declared a as an int and iterated 31 times, you would see undefined behavior corresponding to a "true" overflow. However, in actual fact, because the below program as written is implementation-defined, on some systems it may in fact raise an arithmetic signal terminating the application at the point of the lossy conversion.

dwightguth:
look okay other than the paragraph I mentioned.

dwightguth:
okay let's merge this if the tests pass.

dwightguth:
Jenkins: test this please

dwightguth:
Not really sure what happened here. seems to be some kind of issue with the dependencies in the makefile relating to kdep, but I inspected the file on disk and it looked okay. So I'm not really sure. It shouldn't have anything to do with this diff either. My guess is some kind of transient race or something. If it recurs I'll look into it otherwise our development efforts are probably better spent elsewhere for the time being unless this starts to get more common.

dwightguth:
Jenkins: test this please

PR #206, opened 2016-01-05
Update Error_Codes.csv

dwightguth:
This has actually already been fixed in the code so this should be the only place left to deal with.


dwightguth:
@grosu please review

PR #210, opened 2016-01-06
remove bell character from escape map because K doesn't have a \a escape

dwightguth:
@chathhorn please review


chathhorn:
Looks good.

PR #217, opened 2016-01-24
Test for bug reported in #211.

### LINKED ISSUE #211: Fix Robby's program
This program should be undefined:

```
 #include <stdio.h>

 unsigned long f() {
  double d = -0.5;
  return d * (1 << 30);
 }
 int main() {
  printf("0x%lx\n",f());
 }
```

chathhorn:
Fixes #211.


dwightguth:
Note that this will still not be live in the version of rv-match we use on the website until I finish updating everything, but the version on Jenkins already works, as you can see.

PR #225, opened 2016-02-09
try to fix race AGAIN

dwightguth:
@chathhorn please review

Not sure this is good enough but it's worth a shot


chathhorn:
Looks good.

PR #227, opened 2016-02-11
composite types need not have the same name for function parameters

dwightguth:
@chathhorn please review

Fixes runtimeverification/rv-match#95


chathhorn:
Looks good.

PR #230, opened 2016-02-18
Stack size checking for RV-Match, part 1

dwightguth:
Basically handles two things: computation of the size of the stack at each point in the execution of the program, and some fixes for large writes to make them not be n^2 due to use of cons list internally for the List sort.

@chathhorn please review. Should be merged together with the second PR


chathhorn:
Looks good.

PR #231, opened 2016-02-18
fix -DFOO= form of flag in kcc

dwightguth:
@chathhorn please review

Fixes runtimeverification/rv-match#101


chathhorn:
Looks good.

PR #232, opened 2016-02-18
remove -Werror from error-recovering profiles when calling gcc -E

dwightguth:
This should allow gcc extensions to be used in preprocessing unless error recovery is turned off.

@chathhorn please review. Merges together with second half of PR.


chathhorn:
Looks good.

PR #235, opened 2016-02-19
Refactoring builtins interface.

chathhorn:
@dwightguth please review.

### DIFF
--- a/semantics/library/bootstrap.k
+++ b/semantics/library/bootstrap.k
@@ -82,26 +82,30 @@ module LIBC-BOOTSTRAP
                hrItem(A1:RValue) hs:: hrItem(A2:RValue))
           => builtin(F, A1, A2)
           [structural]
-     rule prepareBuiltin(Identifier(F:String),
-               hrItem(A1:RValue) hs:: hrItem(A2:RValue) hs:: hrItem(A3:RValue))
+     rule prepareBuiltin(Identifier(F:String), hrItem(A1:RValue)
+               hs:: hrItem(A2:RValue) hs:: hrItem(A3:RValue))
           => builtin(F, A1, A2, A3)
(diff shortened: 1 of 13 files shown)


dwightguth:
the native builtins can have any number of arguments, so you need to set up a fallthrough case for handling builtins with greater than 6 arguments, and make sure any builtins that are capable of having greater than 6 arguments can go through that functionality.

dwightguth:
other than that, it looks good though.

semantics/library/bootstrap.k
```diff
@@ -49,14 +58,50 @@ module LIBC-BOOTSTRAP
...
+               hrItem(A1:RValue) hs:: hrItem(A2:RValue) hs:: hrItem(A3:RValue) hs:: hrItem(A4:RValue) hs:: hrItem(A5:RValue))
+          => builtin(F, A1, A2, A3, A4, A5)
+          [structural]
+     rule prepareBuiltin(Identifier(F:String),
+               hrItem(A1:RValue) hs:: hrItem(A2:RValue) hs:: hrItem(A3:RValue)
+               hs:: hrItem(A4:RValue) hs:: hrItem(A5:RValue) hs:: hrItem(A6:RValue))
+          => builtin(F, A1, A2, A3, A4, A5, A6)
+          [structural]
+     rule prepareBuiltin(Identifier(F:String), Args:HeatList)
+          => builtin*(F, hListToList(Args))
```
dwightguth: I think you're going to see a confluence error here because there is a critical pair between this rule and the above rules. You probably want to assert something about the length of the list in the side condition here...

chathhorn: Oops.

chathhorn:
Jenkins: test this please

PR #236, opened 2016-02-19
fix build on windows by ensuring that we resolve symlink instead of c…

dwightguth:
…opying it

@chathhorn please review


chathhorn:
Looks good.

PR #237, opened 2016-02-19
fixes to stack on x86_64 profile

dwightguth:
There were a couple issues with the last commit that slipped through. I think this should work and not duplicate any local variables this time, as well as handling incomplete declarations and VLA's correctly.

@chathhorn please review. See part 2 also.


chathhorn:
Looks good.

PR #238, opened 2016-02-22
Varargs

chathhorn:
@dwightguth please review.

### DIFF
--- a/semantics/language/common/builtin.k
+++ b/semantics/language/common/builtin.k
@@ -8,6 +8,11 @@ module C-BUILTIN-SYNTAX
                   | "extensionBuiltins" [function]
      syntax KItem ::= "handleBuiltin"
                     | handleBuiltinConstant(CId, Type)
+
+     syntax Type ::= "mtx_t" [function]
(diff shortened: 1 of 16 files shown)


dwightguth:
looks okay, but shouldn't we be deleting incSymbolic, varg, and vargResult if they're no longer being used?

semantics/language/common/builtin.k
```diff
@@ -71,12 +72,14 @@ module C-BUILTIN
...
+          "__va_inc" |-> fun(ptr(void), va_list)
+          "__va_start" |-> fun(void, ptr(va_list), ptr(void))
           extensionBuiltins
 
-     syntax Type ::= "mtx_t" [function]
      rule mtx_t => struct("__typeref_mtx_t")
-
-     syntax Type ::= "jmp_buf" [function]
      rule jmp_buf => arr(struct("__typeref_jmp_buf"), 1)
+     rule va_list => t(.Set, cfg:ptrdiffut)
```
dwightguth: is there a reason you made this ptrdiff_t and not another type?

PR #241, opened 2016-02-24
Fix for issue with building sqlite3

msaxena2:
WIP, not yet ready. Fixes issue that occurs while building sqlite3 with rv-match, dealing with type of assignment expression.


dwightguth:
@chathhorn can you review this once @msaxena2 says it's ready, since I basically worked very closely with him on it?

msaxena2:
@chathhorn @dwightguth Please review.

dwightguth:
looks good, but I want @chathhorn to take a look also

chathhorn:
So I think this needs to be done basically anytime you're performing an lvalue conversion (but it isn't going through the normal channels, i.e. commom/expr/eval.k). With that in mind, I think you're missing cases and you're also doing it in places where it isn't needed (i.e., stripping qualifiers on the right operand, which is already lvalue-converted).

semantics/language/execution/expr/assignment.k
```diff
@@ -39,7 +39,7 @@ module C-EXPR-ASSIGNMENT
           ~> checkArrayStaticAssign(T', V)
           ~> checkRestrictAssign(getRestrictBlock(T), getRelevantTags(Loc, V))
           ~> write(Loc, V, T')
-          ~> transferProv(Loc, tv(V, T))
+          ~> transferProv(Loc, tv(V, stripQualifiers(T)))
           requires isRestrictType(T)
```
chathhorn: The changes above shouldn't be needed, I don't think, or else the strip should happen sooner (when the actual lvalue conversion occurs).

PR #245, opened 2016-02-25
use map for configuration

dwightguth:
@chathhorn please review

Don't merge this yet, since it is dependent on code fixes in other pull requests. At a high level, we are making the tu cell be a map whose key is the first child (ie tu-id). This makes looking up info about a particular tu happen in log time instead of linear time.


dwightguth:
Jenkins: test this please

chathhorn:
Looks good.

PR #252, opened 2016-03-04
Refactoring of qualified types.

chathhorn:
This is meant to make a cleaner distinction between unqualifed and qualified types, mostly by introducing a production (ut) for types without qualifiers.

@dwightguth please review.


chathhorn:
Jenkins: test this please.

chathhorn:
Jenkins: test this please.

pdaian:
@chathhorn looks like it hit a timeout, because I was using that machine for other things at the same time.  Perhaps retest (not sure if I have perms, but let's try it: 
Jenkins: test this please. )?

I've increased the timeout from 20 mins to 30 to compensate for other projects potentially building on the machine (build takes 15 on an idle Jenkins, so timeout should definitely be above 30).

pdaian:
Hmm, still timing out.  Let's try again, sorry about that:

Jenkins: test this please

chathhorn:
It's probably something I did, I was having performance issues with this stuff (although it's a bit frustrating that my fixes seem to have made it worse).

chathhorn:
It looks like it's just spinning at the very end until it times out. Is there still some process running or something?

chathhorn:
Moving to rv-match repo.

PR #262, opened 2017-01-19
C++ enumerations.

h0nzZik:
Not finished yet. @dwightguth can you please give me some feedback?


semantics/cpp14/language/translation/operators.k
```diff
@@ -26,6 +26,9 @@ module CPP-OPERATORS
           => builtinOperator(O, E1)
           requires notBool isCPPClassType(T1) andBool notBool isCPPEnumType(T1)
 
+     rule #UnaryOperator(operator&, E1::Expr, T1::CPPEnumType)
```
dwightguth: This rule is not correct. In the case where T1 is an enumeration type, `#UnaryOperator` needs to perform overload resolution to determine whether an overloaded operator exists. This happens by computing the candidate sets according to paragraph 13.3.1.2:3. You don't have to implement the full generality of operator overloading, however, you need to do at least the following:

1. Compute the set of member candidates to be empty because T1 is not a class type.
2. Perform unqualified name lookup on the operator in order to compute the candidate set of non-member candidates. This needs to include arg dependent name lookup as well.
3. Compute the set of built-in candidates for the unary operator & as the empty set.
4. If all three candidate sets are empty, **then** you can select builtinOperator(operator&, E1).

semantics/cpp14/language/translation/syntax.k
```diff
@@ -79,9 +79,11 @@ module CPP-ABSTRACT-SYNTAX
                    | TemplateInstantiationDefinition(AType, Decl)
                    | PartialSpecialization(List, List, Decl)
 
+     syntax Scoped ::= Scoped() | Unscoped()
+
      syntax Decl ::= ClassDef(Tag, CId, NNS, List, List)
                    | TypeDecl(AType)
-                   | EnumDef(CId, NNS, List)
+                   | EnumDef(CId, NNS, Scoped, CPPType, List) [strict(4)]
```
dwightguth: This should be an AType, not a CPPType.

semantics/cpp14/language/translation/typing/expr.k
```diff
@@ -119,6 +119,11 @@ module CPP-TYPING-EXPR
           => typeofBuiltinOperator(O, T1)
           requires notBool isCPPClassType(T1) andBool notBool isCPPEnumType(T1)
 
+     rule typeofOperator(operator&, T::CPPEnumType)
```
dwightguth: operator overloading resolution has to happen here too.

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -0,0 +1,55 @@
...
+          <curr-tu> Tu::String </curr-tu>
+          <tu-id> Tu </tu-id>
+          (.Bag => <cppenum>...
+               <enum-id> I </enum-id>
+               <enum-ut> UT </enum-ut>
+               <scoped> S </scoped>
+           ...</cppenum>)
+
+
+     rule <k> EnumDef(X::CId, NoNNS(), S::Scoped, UT::CPPType, Enumerators::List) =>
```
h0nzZik: Yes, I haven't implemented Enumerators yet. That is also the reason why I provided only one simple test file.

semantics/common/configuration.k
```diff
@@ -101,6 +101,18 @@ module COMMON-CONFIGURATION
                          <access-specifier> .K </access-specifier>
                     </class>
                </classes>
+
+               <cpp-enums>
+                    <cppenum multiplicity="*" type="Map">
+                         <enum-id> 0 </enum-id>
```
h0nzZik: But what about unnamed enums (and classes)? We are able to decltype() an instance of anonymous enum/class.

semantics/cpp14/language/translation/decl/enum.k
dwightguth: alright. I'm okay if we defer that to a later pull request, but if we do that can we change the matching on a variable to matching on an empty list so that the code gets stuck if it reaches a point where it needs a feature we haven't implemented yet?

semantics/common/configuration.k
dwightguth: ah, yes. This is sort of an ongoing issue. There is actually a bug in the semantics dealing with anonymous enums and classes at the moment. I would prefer to solve that issue separately from the issue from the enums, but if you want to tackle it now, what I would suggest is to give all anonymous classes and enums a numeric id as part of their CId. Like, right now ClangKast.cc parses the name of an anonymous struct as `#NoName`. I would propose to change that to `#NoName(Int)`. But it has to happen in the parser before the semantics even starts or the code will not behave correctly no matter what you do.

dwightguth: What this involves is a change to TraverseDeclarationName and TraverseIdentifierInfo if you're curoius.

h0nzZik: Would it be ok to keep the `<enum-id>` version until we use the `#NoName(Int)` solution?

dwightguth: I would prefer to keep things consistent. Right now the keys for classes and namespaces are identifiers, having the keys for enums be integers introduces an inconsistency in the design. Is there a specific program you're trying to run that doesn't execute correctly with that solution?

h0nzZik: No, I am trying only toy examples. Ok, I will use names.

h0nzZik:
I think I have fixed those issues. I will implement operators later. Enumerators are in progress in separate branch.

dwightguth:
So, I'm not really comfortable with the stuff specific to unary & because it's technically an incorrect rule being added to the semantics. But I'm okay with this being merged if you were to remove those rules and implement them later.

h0nzZik:
You are right about the operator. I intended to remove the rule but forgot to do so.

dwightguth:
so, I would prefer in the future if you wouldn't rebase pull requests after they've been approved and before they've been merged. I now have to re-review everything I already reviewed, which is going to be a pain, and if you keep doing that then I'm just going to have to ignore your work in the future as it will turn into a waste of developer resources.

h0nzZik:
I understand - sorry for wasting your time.

h0nzZik:
Replaced by https://github.com/kframework/c-semantics/pull/271.

PR #265, opened 2017-02-26
name not found ==> ill-formed

h0nzZik:
When name is not found during translation phase, `notFound(X)` term does not always make compilation fail, which is probably wrong. I am not sure exactly what message to pass to term ILL(_, _). Note that the following rule does not really help:

`rule resolveUniqueDecl(notFound(_), _) => ILL("", "name not found")`

### DIFF
--- a/semantics/cpp14/language/translation/overloading.k
+++ b/semantics/cpp14/language/translation/overloading.k
@@ -22,6 +22,7 @@ module CPP-OVERLOADING
      imports CPP-OVERLOADING-SYNTAX
      imports C-CONFIGURATION
      imports INT
+     imports STRING
      imports K-EQUAL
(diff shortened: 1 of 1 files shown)


dwightguth:
the rule should be 
```
rule (.K => ILL("TOL1", "No declaration found for name '" +String idToString(X) +String "'.")) ~> resolveUniqueDecl(notFound(X::CId), _)
```

dwightguth:
You may also need to add a side condition to the rule immediately above, something to the extent of `andBool notBool isNotFoundNameRef(V)`

dwightguth:
I'm closing this since it was merged in the upstream private repository. You should see it get pushed down into the public repo sometime tonight.

PR #266, opened 2017-02-28
Value category for sizeof()

h0nzZik:
Value category for sizeof() expression was missing, so it was not possible to compile code like the one in the test.


h0nzZik:
Merged in 6430f9a9e0ad7c8ac97b57fbe3abe0bd1c3b1ad3.

PR #270, opened 2017-03-06
Compile-time evaluation of boolean expr

h0nzZik:
Adds an ability to evaluate boolean expressions to static semantics. For example, the following piece of code failed to compile due to missing rule for negation.

```
if (!true)
    ;
```

### DIFF
--- a/semantics/cpp14/language/translation/expr/logical.k
+++ b/semantics/cpp14/language/translation/expr/logical.k
@@ -21,6 +21,11 @@ module CPP-EXPR-LOGICAL
      rule ! (V:PRVal => convertType(type(bool), V))
           requires notBool isCPPBoolType(type(V))
 
+     rule ! prv(0, Tr::Trace, T:CPPBoolType)
+          => prv(1, combine(Tr, #klabel(`!__CPP-SYNTAX`)), T)
+     rule ! prv(1, Tr::Trace, T:CPPBoolType)
+          => prv(0, combine(Tr, #klabel(`!__CPP-SYNTAX`)), T)
+
      context (HOLE:Expr => reval(HOLE)) || _ [result(PRVal)]
      context _ || (HOLE:Expr => reval(HOLE)) [result(PRVal)]
      context (HOLE:Expr => reval(HOLE)) && _ [result(PRVal)]


dwightguth:
This seems like a strange solution to the problem you describe... We maintain separate semantics for && and || because they have different evaluation strategies at compile time versus runtime. I would prefer to keep it that way because looking at this diff I can't predict what it will actually do to the semantics. Can you please tell me exactly what configuration got stuck at translation time in the above program? I'm pretty sure there's a better solution out there.

h0nzZik:
The top of k cell looks like this:
```
`!__CPP-SYNTAX`(prv(
	#token("1","Int"),
	hasTrace(`BoolLiteral`(#token("true","Bool"))),
	tcpp(quals(`.Set`(.KList)),`.Set`(.KList),`bool_CPP-TYPING-SYNTAX`(.KList))
))~>
`#freezer#ExprLoc0`(`CabsLoc`(...))~>
#freezerrevalCpp0(.KList)~>
#freezerIfStmt2(BlockStmt`(...),`BlockStmt`(...))~>
```
It is sufficient to add following rules to `translation/expr/logical.k`:
```
rule ! prv(0, _, T:CPPBoolType)
    => prv(1, noTrace, T)
rule ! prv(1, _, T:CPPBoolType)
    => prv(0, noTrace, T)
```
and it works for given example. If you agree, I will commit those rules together with a test, and remove the current commit.

dwightguth:
I believe what you actually want is the following rules added to translation/expr/logical.k:

```
rule ! prv(0, Tr::Trace, T:CPPBoolType)
     => prv(1, combine(Tr, #klabel(`!__CPP-SYNTAX`)), T)
rule ! prv(1, Tr::Trace, T:CPPBoolType)
     => prv(0, combine(Tr, #klabel(`!__CPP-SYNTAX`)), T)
```

But yes, adding them to translation/expr/logical is the preferred solution.

dwightguth:
There are four types of casts in K:

`V:Sort` tells the compiler that V is a variable of sort Sort and that this constraint should be checked at runtime; ie, it causes the rule it is in to gain a side condition that causes the rule not to apply if the value matched for V is not of sort Sort.

`V::Sort` tells the compiler that V should be assumed to be of sort Sort during parsing, but that no constraints should be checked at runtime. This can lead to poorly-typed configurations if it's abused, but it should be used whenever possible in cases when the programmer is confident that the value is guaranteed to be of that sort no matter what, because it improves performance of the interpreter.

`{V}:>Sort` Functions similar to a cast in other languages. It tells the compiler that something which is statically inferred elsewhere as another sort should be treated here as if it is of sort Sort. This is useful for expressions such as `isCPPFunctionType(T) andBool isAdjustedType({T}:>CPPFunctionType)` where you are checking the sort of a variable explicitly rather than with the `T:CPPFunctionType` syntax.

Finally, `{V}<:Sort` Tells the parser that V should be assumed to be of sort Sort during parsing, similar to `V::Sort`, but the distinction is that it creates an expression of sort K, which can appear anywhere in a term. This is used only in very rare circumstances in order to perform disambiguation between conflicting syntax productions. I do not anticipate that you should need to use it, but I explain here in the interests of completeness.

PR #271, opened 2017-03-11
C++ Enums + enumerators

h0nzZik:
Implements: enums, enumerators, their underlying type, lookup.

Does not implement: enums in functions/classes, std::underlying_type, opaque enums.

Depends on:
https://github.com/kframework/c-semantics/pull/270.
https://github.com/kframework/c-semantics/pull/268

Closes https://github.com/kframework/c-semantics/pull/262

### DIFF
--- a/semantics/cpp14/language/common/conversion.k
+++ b/semantics/cpp14/language/common/conversion.k
@@ -86,9 +86,9 @@ module CPP-CONVERSION
           ~> convertTypeHold(T, prv(C, Tr, T'))
           requires isCPPSignedType(underlyingType(T)) andBool notBool inRange(C, T)
      rule convertType(T::CPPType, prv(C::CPPValue, Tr::Trace, T':CPPIntegerType)) => arithInterpret(T, C, Tr)
-          requires isCPPUnsignedType(underlyingType(T)) andBool inRange(C, T)
+          requires notBool isCPPEnumType(T) andBool isCPPUnsignedType(underlyingType(T)) andBool inRange(C, T)
      rule convertType(T::CPPType, prv(N:Int, Tr::Trace, T':CPPIntegerType)) => prv(N modInt (max(T) +Int 1), Tr, T)
-          requires isCPPUnsignedType(underlyingType(T)) andBool notBool inRange(N, T)
+          requires notBool isCPPEnumType(T) andBool isCPPUnsignedType(underlyingType(T)) andBool notBool inRange(N, T)
      rule convertType(T:CPPIntegerType, prv(C::CPPValue, Tr::Trace, T':CPPBoolType)) => prv(C, Tr, T)
      rule convertType(T:CPPBoolType, prv(0, Tr::Trace, T':CPPIntegerType)) => prv(0, Tr, T)
      rule convertType(T:CPPBoolType, prv(I:Int, Tr::Trace, T':CPPIntegerType)) => prv(1, Tr, T)
@@ -230,7 +230,25 @@ module CPP-CONVERSION
                                                 type(underlyingType(T)) 
                                                 #fi #fi #fi #fi #fi #fi
           requires T ==K char16_t orBool T ==K char32_t orBool T ==K wchar_t
-     // TODO(dwightguth): 4.5:3-5
+
+     // 4.5:3
+     // TODO(h0nzZik) (1) store the type in configuration for performance reasons,
+     //               (2) what if we have an extended integer type other then the underlying type?
+     rule promote(t(_, _, unscopedEnum(_, _, false) #as T::CPPSimpleType)) =>
+               #if broaderTypeThan(type(int), type(T)) #then type(int) #else
+               #if broaderTypeThan(type(unsigned), type(T)) #then type(unsigned) #else
+               #if broaderTypeThan(type(long), type(T)) #then type(long) #else
+               #if broaderTypeThan(type(unsigned-long), type(T)) #then type(unsigned-long) #else
+               #if broaderTypeThan(type(long-long), type(T)) #then type(long-long) #else
+               #if broaderTypeThan(type(unsigned-long-long), type(T)) #then type(unsigned-long-long) #else
+               type(underlyingType(T))
+               #fi #fi #fi #fi #fi #fi
+
+     // 4.5:4
+     rule promote(t(_, _, unscopedEnum(_, UT::CPPIntegerType, true))) => promote(UT)
+
+     // TODO(dwightguth): 4.5:5
+
      // 4.5:6
      rule promote(t(_, _, bool)) => type(int)
      rule promote(T::CPPType) => T
--- a/semantics/cpp14/language/common/conversion.k
+++ b/semantics/cpp14/language/common/conversion.k
@@ -241,7 +241,7 @@ module CPP-CONVERSION
                #if broaderTypeThan(type(unsigned-long), type(T)) #then type(unsigned-long) #else
                #if broaderTypeThan(type(long-long), type(T)) #then type(long-long) #else
                #if broaderTypeThan(type(unsigned-long-long), type(T)) #then type(unsigned-long-long) #else
-               type(underlyingType(T))
+               type(underlyingType(T)) // Not reachable without extended integer types
                #fi #fi #fi #fi #fi #fi
 
      // 4.5:4
--- a/semantics/cpp14/language/common/memory/reading.k
+++ b/semantics/cpp14/language/common/memory/reading.k
@@ -47,4 +47,8 @@ module CPP-MEMORY-READING
      rule interpret(prv(0, Tr::Trace, T:CPPNullPtrTType)) => prv(nullptrVal, Tr, T)
      rule interpret(prv(0, Tr::Trace, T:CPPPointerType)) => prv(NullPointer, Tr, T)
 
+     syntax PRVal ::= interpretEnum(PRVal, CPPEnumType)
+     rule interpret(prv(I:Int, Tr::Trace, T:CPPEnumType)) => interpretEnum(interpret(prv(I, Tr, underlyingType(T))), T)
+     rule interpretEnum(prv(I:Int, Tr::Trace, T:CPPIntegerType), E:CPPEnumType) => prv(I, Tr, E)
+
 endmodule
--- a/semantics/cpp14/language/common/memory/reading.k
+++ b/semantics/cpp14/language/common/memory/reading.k
@@ -47,8 +47,6 @@ module CPP-MEMORY-READING
      rule interpret(prv(0, Tr::Trace, T:CPPNullPtrTType)) => prv(nullptrVal, Tr, T)
      rule interpret(prv(0, Tr::Trace, T:CPPPointerType)) => prv(NullPointer, Tr, T)
 
-     syntax PRVal ::= interpretEnum(PRVal, CPPEnumType)
-     rule interpret(prv(I:Int, Tr::Trace, T:CPPEnumType)) => interpretEnum(interpret(prv(I, Tr, underlyingType(T))), T)
-     rule interpretEnum(prv(I:Int, Tr::Trace, T:CPPIntegerType), E:CPPEnumType) => prv(I, Tr, E)
+     rule interpret(prv(I:Int, Tr::Trace, T:CPPEnumType)) => #fun(prv(I::Int, Tr::Trace, (T => T)))(interpret(prv(I, Tr, underlyingType(T))))
 
 endmodule
--- a/semantics/cpp14/language/common/typing.k
+++ b/semantics/cpp14/language/common/typing.k
@@ -111,7 +111,7 @@ module CPP-TYPING-SYNTAX
      syntax CPPSimpleArrayTypeExpr ::= dynamicArrayType(CPPType, Expr)
 
      syntax CPPSimpleScopedEnumType ::= scopedEnum(id: EnumId, underlyingType: CPPIntegerType)
-     syntax CPPSimpleUnscopedEnumType ::= unscopedEnum(id: EnumId, underlyingType: CPPIntegerType)
+     syntax CPPSimpleUnscopedEnumType ::= unscopedEnum(id: EnumId, underlyingType: CPPIntegerType, utFixed: Bool)
      syntax CPPSimpleEnumType ::= CPPSimpleScopedEnumType | CPPSimpleUnscopedEnumType
      syntax CPPSimpleCompoundType ::= CPPSimpleEnumType
 
@@ -426,7 +426,7 @@ module CPP-TYPING
 
      rule t(Q::Quals, _, classType(N1::Namespace :: Class(T1::Tag, X1::CId, Ts1::TemplateArgs))) ==Type t(Q::Quals, _, classType(N2::Namespace :: Class(T2::Tag, X2::CId, Ts2::TemplateArgs))) => N1 ==K N2 andBool T1 ==K T2 andBool X1 ==K X2 andBool Ts1 ==Types Ts2
      rule t(Q::Quals, _, scopedEnum(X::EnumId, _)) ==Type t(Q, _, scopedEnum(X::EnumId, _)) => true
-     rule t(Q::Quals, _, unscopedEnum(X::EnumId, _)) ==Type t(Q, _, unscopedEnum(X::EnumId, _)) => true
+     rule t(Q::Quals, _, unscopedEnum(X::EnumId, _, _)) ==Type t(Q, _, unscopedEnum(X::EnumId, _, _)) => true
 
      syntax Bool ::= TemplateArgs "==Types" TemplateArgs [function]
      rule T1::CPPType, Ts1::TemplateArgs ==Types T2::CPPType, Ts2::TemplateArgs => T1 ==Type T2 andBool Ts1 ==Types Ts2
@@ -440,7 +440,7 @@ module CPP-TYPING
      rule underlyingType(t(Q::Quals, Mods::Set, T::CPPSimpleType))
           => t(Q, Mods, underlyingType(T))
      rule underlyingType(scopedEnum(_, T::CPPIntegerType)) => simpleType(T)
-     rule underlyingType(unscopedEnum(_, T::CPPIntegerType)) => simpleType(T)
+     rule underlyingType(unscopedEnum(_, T::CPPIntegerType, _)) => simpleType(T)
 
 
      rule inRange(I:Int, T::CPPIntegerType)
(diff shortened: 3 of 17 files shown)


semantics/cpp14/language/common/conversion.k
```diff
@@ -230,7 +230,25 @@ module CPP-CONVERSION
...
+     // TODO(h0nzZik) (1) store the type in configuration for performance reasons,
+     //               (2) what if we have an extended integer type other then the underlying type?
+     rule promote(t(_, _, unscopedEnum(_, _, false) #as T::CPPSimpleType)) =>
+               #if broaderTypeThan(type(int), type(T)) #then type(int) #else
+               #if broaderTypeThan(type(unsigned), type(T)) #then type(unsigned) #else
+               #if broaderTypeThan(type(long), type(T)) #then type(long) #else
+               #if broaderTypeThan(type(unsigned-long), type(T)) #then type(unsigned-long) #else
+               #if broaderTypeThan(type(long-long), type(T)) #then type(long-long) #else
+               #if broaderTypeThan(type(unsigned-long-long), type(T)) #then type(unsigned-long-long) #else
+               type(underlyingType(T))
```
dwightguth: This case should not actually be reachable without any extended integer types, and I would prefer a comment to that effect.

semantics/cpp14/language/common/memory/reading.k
```diff
@@ -47,4 +47,8 @@ module CPP-MEMORY-READING
      rule interpret(prv(0, Tr::Trace, T:CPPNullPtrTType)) => prv(nullptrVal, Tr, T)
      rule interpret(prv(0, Tr::Trace, T:CPPPointerType)) => prv(NullPointer, Tr, T)
 
+     syntax PRVal ::= interpretEnum(PRVal, CPPEnumType)
+     rule interpret(prv(I:Int, Tr::Trace, T:CPPEnumType)) => interpretEnum(interpret(prv(I, Tr, underlyingType(T))), T)
```
dwightguth: this is a good candidate for a lambda application. You should be able to rewrite the rule like so:
```
rule interpret(prv(I:Int, Tr::Trace, T:CPPEnumType)) => #fun(prv(I::Int, Tr::Trace, (_ => T)))(interpret(prv(I, Tr, underlyingType(T))))
```

semantics/cpp14/language/common/typing.k
```diff
@@ -448,6 +474,12 @@ module CPP-TYPING
      rule min(T:CPPWideCharType) => min(underlyingType(T))
      rule max(T:CPPWideCharType) => max(underlyingType(T))
 
+     rule min(t(_, _, scopedEnum(_, UT::CPPIntegerType))) => min(UT)
+     rule max(t(_, _, scopedEnum(_, UT::CPPIntegerType))) => max(UT)
+
+     rule min(t(_, _, unscopedEnum(E::EnumId, _, _))) => #getEnumMin(#getEnumInfo(E, #configuration))
```
dwightguth: this doesn't seem quite right. In the case of an unscoped enum whose type is fixed, it should have the same values as the scoped enum above.

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -0,0 +1,218 @@
...
+          (.Bag => <cppenum>...
+               <enum-id> getEnumId(E) </enum-id>
+               <enum-type> E </enum-type>
+               <scoped> false </scoped>
+               <enum-complete> false </enum-complete>
+           ...</cppenum>)
+
+     syntax KItem ::= declareEnumerators(CPPEnumType, min: Int, max: Int, next: Int, nextType: CPPSimpleType, List, fixed: Bool, first: Bool)
+
+     rule declareEnumerators(_, _, _, Next::Int, NextType::CPPSimpleType, ListItem(Enumerator(_, NoExpression() => prv(Next, noTrace, type(NextType)))) _::List, _, _)
```
dwightguth: I would prefer that this case generate an UNSPEC warning if first is equal to true.

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -0,0 +1,218 @@
...
+     context declareEnumerators(_, _, _, _, _, ListItem(Enumerator(_, HOLE:Expr)) _::List, _, _)
+
+     // Type of an enumerator inside the declaration is enum's underlying type
+     // if it is fixed, and type of initializing expression otherwise,
+     // Hence we need to store the type of underlying expression...
+     rule declareEnumerators(T::CPPEnumType, Min::Int, Max::Int, Next::Int, _, ListItem(Enumerator(X::CId, prv(V::Int, _, t(_, _, EType::CPPSimpleType)) #as Pr::PRVal)) L::List, Fixed::Bool, true)
+          => declareEnumerator(getEnumId(T), X, Pr, Fixed, underlyingType(T))
```
dwightguth: this does not seem to be correctly computing the type of the enumerator in the case when EType is an unscoped enumeration type.

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -0,0 +1,218 @@
...
+          ~> declareEnumerators(T, V, V, V +Int 1, chooseEnumeratorType(EType, V +Int 1), L, Fixed, false)
+
+     rule declareEnumerators(T::CPPEnumType, Min::Int, Max::Int, Next::Int, _, ListItem(Enumerator(X::CId, prv(V::Int, _, t(_, _, EType::CPPSimpleType)) #as Pr::PRVal)) L::List, Fixed::Bool, false)
```
dwightguth: I would prefer that instead of calling chooseEnumerationType we add a side condition that V +Int 1 is in the range of the type of the preceding enumerator, and if it isn't, then we generate an UNSPEC warning

semantics/cpp14/language/translation/decl/enum.k
h0nzZik: There is no UNSPEC in C++ semantic, so I will take the definition from C semantics. I guess that I should do the rewrite even if it is unspecified, so that rv-match could recover from that. So the rule will look like
`rule <k> (.K => UNSPEC(...)) ~> a(b=>c, true)` ... </k>

h0nzZik: Also I presume that error titles are abbreviations of paths to source files, e.g. `CCV3` means `common/conversion.k` - but correct me if I am wrong.

dwightguth: Yes, it should be easy to add a rule for UNSPEC to cpp14/language/common/error.k. You are correct about what the letters in the error code refer to as well. The number just uniquely identifies the error within the file. Note that you don't need to worry about collisions between C++ and C error codes, the semantics automatically handles this.

h0nzZik: Just to clarify: the unspecified behavior you are referring to is in that the type of first enumerator is unspecified (7.2:5.2)? It is the case only for enums whose underlying type is not fixed. Also, it is quite common to see code like
```C++
enum E {A, B, C};
```
and I am afraid that UNSPEC would stop the compilation of this code.

dwightguth: After discussions with my boss, we've concluded that the right way to proceed here is that I'm going to spend some time tomorrow adding a rudimentary system for reporting errors and recovering from them back into the open source C semantics. As a result of this change you should be able to successfully execute programs which generate the UNSPEC behavior. All you should need to do on your end is to create a rule like follows:

rule (.K => UNSPEC(...)) ~> (Pattern => recoverFromError(Expression))

Where Pattern matches the unspecified behavior and Expression represents the code which should replace Pattern at the top of the k cell after the error has been reported.

So yes, go ahead and add the UNSPEC in this case.

semantics/cpp14/language/translation/decl/enum.k
h0nzZik: We can generate an UNSPEC warning. but still have to choose some sufficiently large type.

dwightguth: Yes, I'm fine with the existing behavior in this case.

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -0,0 +1,218 @@
...
+             ListItem(short)
+             ListItem(unsigned-short)
+             integralTypesGreaterThanInt
+
+
+     
+     syntax CPPSimpleType ::= firstSuitableType(Int, Int, List) [function]
+     rule firstSuitableType(Min::Int, Max::Int, ListItem(T::CPPSimpleType) _) => T
+          requires Min >=Int min(type(T)) andBool Max <=Int max(type(T))
+     rule firstSuitableType(_, _, .List) => int
```
dwightguth: isn't the program ill-formed in this case, based on the standard?

h0nzZik: Yes, 7.2:5.3 says it.

dwightguth: if we don't expect this case to ever actually occur in well-formed programs, I would prefer to keep the function partial and have it get stuck if somehow we end up at this case by mistake.

semantics/cpp14/language/translation/name.k
```diff
@@ -268,16 +304,28 @@ module CPP-TRANSLATION-NAME
                     | resolveElabSpecifier(K, StrictList) [strict(1, 2), klabel(resolveElabSpecifier2)]
                     | specializeTypeTemplate(QualId, CPPType, List)
 
-     rule <k> nameLookup(X::CId, Tag::Tag, IsNNS::Bool) => lookupNameInFullClass(X, classFromScope(Scope), isBlockScope(Scope), Tag, IsNNS) ...</k>
+     rule <k> nameLookup(X::CId, Tag::Tag, IsNNS::Bool) => nameLookupInScope(X, Tag, IsNNS, Scope) ...</k>
           <curr-scope> Scope::Scope </curr-scope>
+
+     rule <k> nameLookupInScope(X::CId, Tag::Tag, IsNNS::Bool, Scope::Scope) => lookupNameInFullClass(X, classFromScope(Scope), isBlockScope(Scope), Tag, IsNNS) ...</k>
```
traiansf: It seems you can drop the <k>  cell here

semantics/cpp14/language/translation/name.k
```diff
@@ -268,16 +304,28 @@ module CPP-TRANSLATION-NAME
...
           requires inClassScope(Scope)
 
+     rule nameLookupInScope(X::CId, Tag::Tag, IsNNS::Bool, enumScope(E::EnumId)) => nameLookupInEnumScope(X, Tag, IsNNS, enumScope(E))
+
```
traiansf: You can have a race condition here between the two rules above, as inClassScope can be true for an EnumScope (you added the rule yourself).

traiansf: And this might be problematic if you have an enum in a class, as in the next example, as if the first rule is chosen, the name will be looked up outside the class.
```
constexpr int a = 4;
struct C {
  enum Enum {
    a = 5,
    b = a*2
  };
};

int main() {
  C::Enum e = C::b;
  return e - 10;
}
```

semantics/cpp14/language/common/typing.k
```diff
@@ -477,8 +477,11 @@ module CPP-TYPING
      rule min(t(_, _, scopedEnum(_, UT::CPPIntegerType))) => min(UT)
      rule max(t(_, _, scopedEnum(_, UT::CPPIntegerType))) => max(UT)
 
-     rule min(t(_, _, unscopedEnum(E::EnumId, _, _))) => #getEnumMin(#getEnumInfo(E, #configuration))
-     rule max(t(_, _, unscopedEnum(E::EnumId, _, _))) => #getEnumMax(#getEnumInfo(E, #configuration))
+     rule min(t(_, _, unscopedEnum(_, UT::CPPIntegerType, _))) => min(UT)
```
dwightguth: this looks non confluent to me. You need to match on fixed = true.

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -80,23 +80,52 @@ module CPP-DECL-ENUM
...
-     syntax KItem ::= declareEnumerators(CPPEnumType, min: Int, max: Int, next: Int, nextType: CPPSimpleType, List, fixed: Bool, first: Bool)
-
-     rule declareEnumerators(_, _, _, Next::Int, NextType::CPPSimpleType, ListItem(Enumerator(_, NoExpression() => prv(Next, noTrace, type(NextType)))) _::List, _, _)
+     syntax KItem ::= declareEnumerators(CPPEnumType, min: Int, max: Int, next: Int, prevType: CPPSimpleType, List, fixed: Bool, first: Bool)
+
+     // First enumerator, fixed underlying type.
+     rule declareEnumerators(E::CPPEnumType, _, _, _, _, ListItem(Enumerator(_, NoExpression() => prv(0, noTrace, underlyingType(E)))) _::List, true, true)
+
+     // First enumerator, underlying type not fixed
+     rule (.K => UNSPEC("TDE2", "No initializer specified for the first enumerator")) ~> (
```
dwightguth: Maybe "No initializer specified for the first enumerator, underlying type of enumerator is unspecified"?

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -80,23 +80,52 @@ module CPP-DECL-ENUM
...
-     rule declareEnumerators(T::CPPEnumType, Min::Int, Max::Int, Next::Int, _, ListItem(Enumerator(X::CId, prv(V::Int, _, t(_, _, EType::CPPSimpleType)) #as Pr::PRVal)) L::List, Fixed::Bool, true)
-          => declareEnumerator(getEnumId(T), X, Pr, Fixed, underlyingType(T))
-          ~> declareEnumerators(T, V, V, V +Int 1, chooseEnumeratorType(EType, V +Int 1), L, Fixed, false)
-
-     rule declareEnumerators(T::CPPEnumType, Min::Int, Max::Int, Next::Int, _, ListItem(Enumerator(X::CId, prv(V::Int, _, t(_, _, EType::CPPSimpleType)) #as Pr::PRVal)) L::List, Fixed::Bool, false)
+     rule declareEnumerators(T::CPPEnumType, Min::Int, Max::Int, Next::Int, _, ListItem(Enumerator(X::CId, prv(V::Int, _, t(_, _, EType::CPPSimpleType)) #as Pr::PRVal)) L::List, Fixed::Bool, First::Bool)
           => declareEnumerator(getEnumId(T), X, Pr, Fixed, underlyingType(T))
-          ~> declareEnumerators(T, minInt(Min, V), maxInt(Max, V), V +Int 1, chooseEnumeratorType(EType, V +Int 1), L, Fixed, false)
-
+          ~> declareEnumerators(T, ite(First, V, minInt(Min, V)), ite(First, V, maxInt(Max, V)), V +Int 1, EType, L, Fixed, false)
```
dwightguth: we actually already have a construct in K for if then else: 
```
#if First #then V #else minInt(Min, V) #fi
```

semantics/cpp14/language/translation/decl/enum.k
dwightguth: I don't see that this has been fixed? specifically refer to 7.2:6.1. If the initializer has unscoped enumeration type, the enumerator has the underlying type of that type.

semantics/cpp14/language/translation/decl/enum.k
h0nzZik: I do not know exactly why, but when I try
``
 declareEnumerators(T, ite(First, V, minInt(Min, V)), #if First #then V #else maxInt(Max, V) #fi , V +Int 1, EType, L, Fixed, false)
```
the compilation fails with
```
File "realdef.ml", line 26878, characters 1268-1282:
Error: This expression has type Constants.K.t = Constants.K.kitem list
       but an expression was expected of type Constants.K.kitem
```
I have seen this problem in other context, too. For example in rules like the following
```
rule declareEnumerators(E::CPPEnumType, _, _, Next::Int, _, ListItem(Enumerator(_, NoExpression() => prv(Next, noTrace, underlyingType(E)))) _::List, Fixed::Bool, First::Bool)
requires Fixed andBool notBool First
```

Comment (dwightguth) [inline semantics/cpp14/language/translation/decl/enum.k]: ah yes. This is a known issue in the ocaml backend. When using a boolean variable in boolean operators (ie and, or, not, and if/then/else), please explicitly type the boolean variable (i.e. `First:Bool`).

**Comment by h0nzZik** [comment]
I am not able to react directly to the comment, but 7.2:5.1. 

> If the initializer has unscoped enumeration type, the enumerator has the underlying type of that type.
is implemented in https://github.com/kframework/c-semantics/pull/271/files#diff-69a194dd7afe72b812771fbf3e2ba7c5R116

File: semantics/cpp14/language/translation/decl/enum.k [hunk 13]
```diff
@@ -0,0 +1,238 @@
...
+          => recoverFromError(
+          declareEnumerators(E, Min, Max, Next::Int, PrevType, ListItem(Enumerator(X, prv(Next, noTrace, type(firstSuitableType(Next, Next, allIntegralTypes))))) L, false, false)
+          ))
+          requires notBool inRange(Next, type(PrevType)) andBool firstSuitableType(Next, Next, allIntegralTypes) =/=K no-type
+
+     rule (.K => ILL("TDE4", "Incremented enumerator value is not representable in any integer type")) ~>
+          declareEnumerators(_, _, _, Next::Int, PrevType::CPPSimpleType, ListItem(Enumerator(_, NoExpression())) _::List, false, false)
+          requires notBool inRange(Next, type(PrevType)) andBool firstSuitableType(Next, Next, allIntegralTypes) ==K no-type
+
+     rule declareEnumerators(_, _, _, _, _, ListItem(Enumerator(_, prv(_, _, T:CPPEnumType => underlyingType(T)))) _::List, _, _)
```
dwightguth: this should be T:CPPUnscopedEnumType

semantics/cpp14/language/translation/decl/enum.k
```diff
@@ -0,0 +1,238 @@
...
+     context declareEnumerators(_, _, _, _, _, ListItem(Enumerator(_, HOLE:Expr)) _::List, _, _)
+
+     // Type of an enumerator inside the declaration is enum's underlying type
+     // if it is fixed, and type of initializing expression otherwise,
+     // Hence we need to store the type of underlying expression...
+     rule declareEnumerators(T::CPPEnumType, Min::Int, Max::Int, Next::Int, _, ListItem(Enumerator(X::CId, prv(V::Int, _, t(_, _, EType::CPPSimpleType)) #as Pr::PRVal)) L::List, Fixed:Bool, First:Bool)
+          => declareEnumerator(getEnumId(T), X, Pr, Fixed, underlyingType(T))
+          ~> declareEnumerators(T, #if First #then V #else minInt(Min, V) #fi, #if First #then V #else maxInt(Max, V) #fi, V +Int 1, EType, L, Fixed, false)
```
dwightguth: this rule needs a side condition declaring that EType isn't a CPPSimpleUnscopedEnumType.

PR #275, opened 2017-04-06
Builtin operators for enums.

h0nzZik:
Builtin operators for enums, as described in 13.6.

### DIFF
--- a/semantics/cpp14/language/translation/name.k
+++ b/semantics/cpp14/language/translation/name.k
@@ -498,6 +498,7 @@ module CPP-TRANSLATION-NAME
           => resolveEmptyOverload(cSetUnion2(C1, C2), list(ListItem(E1) ListItem(E2)), E1 -> no-template E2, BinaryOperator(operator->, E1, E2))
 
      // 13.6:15
+/*
      rule addBuiltinCandidates(operator<, E1::Expr, T1:CPPEnumType, _, E2::Expr, T2:CPPEnumType, _, C1:CandidateSet, C2:CandidateSet)
           => resolveEmptyOverload(cSetUnion2(C1, C2), list(ListItem(E1) ListItem(E2)), E1 < E2, BinaryOperator(operator<, E1, E2))
           requires simpleType(T1) ==K simpleType(T2)
@@ -521,11 +522,50 @@ module CPP-TRANSLATION-NAME
      rule addBuiltinCandidates(operator!=, E1::Expr, T1:CPPEnumType, _, E2::Expr, T2:CPPEnumType, _, C1:CandidateSet, C2:CandidateSet)
           => resolveEmptyOverload(cSetUnion2(C1, C2), list(ListItem(E1) ListItem(E2)), E1 != E2, BinaryOperator(operator!=, E1, E2))
           requires simpleType(T1) ==K simpleType(T2)
+*/
 
-     // 13.6:20
-     rule addBuiltinCandidates(operator=, E1::Expr, T1:CPPEnumType, _, E2::Expr, T2:CPPEnumType, _, C1:CandidateSet, C2:CandidateSet)
-          => resolveEmptyOverload(cSetUnion2(C1, C2), list(ListItem(E1) ListItem(E2)), E1 := E2, BinaryOperator(operator=, E1, E2))
-          requires simpleType(T1) ==K simpleType(T2)
+
+     syntax Map ::= builtinNonvolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinVolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinNonvolatileArithAssignOperator(Expr, CPPType, ValueCategory, Expr, CPPType, ValueCategory) [function]
+
+     rule builtinNonvolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
+                      toCPPTypes(ListItem(type(lvRefType(type(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                      noMethod, CPPLinkage, exSet(.Set)))
+                |-> builtinOp(E1 := E2))
+
+          requires isCPPEnumType(T1) andBool notBool isVolatileType(T1) andBool viableType(type(simpleType(T1)), T2, VC)
+
+     rule builtinNonvolatileEnumAssignOperator(...) => .Map [owise]
+
+     rule builtinVolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
+                      toCPPTypes(ListItem(type(lvRefType(volatile(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                      noMethod, CPPLinkage, exSet(.Set)))
+                |-> builtinOp(E1 := E2))
+
+          requires isCPPEnumType(T1) andBool isVolatileType(T1) andBool viableType(type(simpleType(T1)), T2, VC)
+
+     rule builtinVolatileEnumAssignOperator(...) => .Map [owise]
+
+     rule builtinNonvolatileArithAssignOperator(E1::Expr, T1::CPPType, VC1::ValueCategory, E2::Expr, T2::CPPType, VC2::ValueCategory) =>
+                   (type(functionType(type(simpleType(T1)),
+                             toCPPTypes(ListItem(type(lvRefType(type(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                             noMethod, CPPLinkage, exSet(.Set))))
+               |-> builtinOp(E1 := E2)
+          requires isCPPArithmeticType(type(simpleType(T1))) andBool viableType(type(lvRefType(type(simpleType(T1)))), T1, VC1) andBool viableType(type(simpleType(T1)), T2, VC2)
+
+     rule builtinNonvolatileArithAssignOperator(...) => .Map [owise]
+
+     // There shouldn't be any user overload with the same parameter list for non-class T1.
+     rule addBuiltinCandidates(operator=, E1::Expr, T1:CPPType, VC1:ValueCategory, E2::Expr, T2:CPPType, VC2:ValueCategory, C1:CandidateSet, C2:CandidateSet)
+          => resolveOverload(cSetUnion2(cSetUnion2(C1, C2), cSet(
+                              builtinNonvolatileEnumAssignOperator(E1, T1, E2, T2, VC2)
+                              builtinVolatileEnumAssignOperator(E1, T1, E2, T2, VC2)
+                              builtinNonvolatileArithAssignOperator(E1, T1, VC1, E2, T2, VC2)
+                              , NoNamespace() :: operator=, .K
+            )), list(ListItem(E1) ListItem(E2)), BinaryOperator(operator=, E1, E2))
 
      syntax Expr ::= resolveEmptyOverload(K, StrictList, Expr, Expr) [strict(1)]
                    | #resolveEmptyOverload(Expr, Expr)
--- a/semantics/cpp14/language/translation/name.k
+++ b/semantics/cpp14/language/translation/name.k
@@ -497,6 +497,36 @@ module CPP-TRANSLATION-NAME
      rule addBuiltinCandidates(operator->, E1::Expr, _, _, E2::Name, _, _, C1:CandidateSet, C2:CandidateSet)
           => resolveEmptyOverload(cSetUnion2(C1, C2), list(ListItem(E1) ListItem(E2)), E1 -> no-template E2, BinaryOperator(operator->, E1, E2))
 
+     // 13.6:15
+     rule addBuiltinCandidates(operator<, E1::Expr, T1:CPPEnumType, _, E2::Expr, T2:CPPEnumType, _, C1:CandidateSet, C2:CandidateSet)
+          => resolveEmptyOverload(cSetUnion2(C1, C2), list(ListItem(E1) ListItem(E2)), E1 < E2, BinaryOperator(operator<, E1, E2))
+          requires simpleType(T1) ==K simpleType(T2)
+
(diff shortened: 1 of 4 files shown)


h0nzZik:
Now I see. So I will use resolveOverload with the union of all three candidate set. The built-in candidate sets will be singletons, because 13.3.1.2:3.3.3 says the operands should be convertible, and enums are not implicitly convertible to anything (except some cv-conversions etc) - is this correct?.

dwightguth:
This is not quite correct. Many of the builtin sets in 13.6 are infinite sets, and therefore it is necessary to prune the search space based on the types of the operands in order to be able to compute the full list of candidates that participate in overload resolution. And indeed, that is the case with the candidate sets in paragraphs 15, 16, 19, and 20, all of which pertain to the operators you defined in this PR. Thus, taking each of these paragraphs individually:

Paragraph 15: both operands must convert to the same pointer or enumeration type. If we assume neither operand is of class type, then this can only occur if one of the arguments is an enumeration. in that case, the enumeration cannot be converted to a pointer, or a different enumeration type, which means it must be the same enumeration type as the builtin candidate. thus there is only one candidate set to consider, namely, the candidate for the enumeration type that is an argument.

Paragraph 16: Here again if we exclude class types then we can ignore this case entirely because one of the arguments must be an enumeration, and that enumeration cannot be converted to a member pointer or to std::nullptr_t.

Paragraph 19: Here if the first argument is the enumeration we must also consider the semantics of reference binding in overloading. However since restrictions on overloading of assignment require that no temporary or user defined conversion applies to the left operand, we are left only with the cases when the initializer expression is the same type as the reference, which it cannot be because the type of the reference is a pointer. If the second argument is the enumeration then we can also ignore this case because enumerations cannot be converted to pointers.

Paragraph 20: Here we can ignore the case of pointers to members for the same reason as paragraph 16. If the first operand is the enumeration, then by the same logic as paragraph 19, the argument will only convert if it's of the exact same enumeration type. And the same is true of the second operand by the same logic as paragraph 15. Thus again, we only need to considre a single case., namely, the candidate for teh enumeration type that is an argument.

Thus as far as these four paragraphs go with infinite sets, you are correct insofar as it concerns operands of non-class type. However, the operators you wish to implement are also governed by paragraphs 12 and 18, and here your logic goes astray. By my count there are quite a number of candidate operator functions that could match enumerations (because unscoped enumerations can be converted to any integer type. However, these candidate sets as described in this paragraph are finite and therefore can be exhaustively enumerated. I would suggest you write code that exhaustively enumerates them and adds the resulting exhaustive set to the builtin candidate set (subject to the filtering done in paragraph 13.3.1.2:3.3.4), and let the overloading mechanism automatically resolve the best viable candidate. You may run into cases where more than one candidate is viable, but in that case I would recommend holding off a couple days because I am in the process of submitting code which will implement the best viable candidate selection, which would narrow it down in most cases to either the exact match or to a promotion.

h0nzZik:
I've tried to implement `operator=` (but forgot about volatile arithmetic LHS - I will fix it later), is this the right way to do it? The code is a bit verbose, so I will refactor it. I would like to implement the other operators in the similar manner.

Also I am not sure how to correctly check whether the left expression can be bound to first parameter, which is a non-const reference; I will probably use something like 
```
viableType(type(lvRefType(type(simpleType(T1)))), T1, VC1)
```
everywhere.

Programs with "bad" assignments (e.g. to `const`) are rejected by `clang-kast`, so testing the semantics against such programs is not easy. Is it important? Could we somehow use `kcc` with hand-crafted AST?

Also, when I tried to use `#if #then #else #fi` instead of `[owise]` rules, I encountered a strange behaviour, when the three maps in `addBuiltinCandidates` rule (resulting from `builtin...AssignOperator`) were not correctly merged - the result was `.Map`, although one of them was not empty.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+*/
+
+
+     syntax Map ::= builtinNonvolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinVolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinNonvolatileArithAssignOperator(Expr, CPPType, ValueCategory, Expr, CPPType, ValueCategory) [function]
+
+     rule builtinNonvolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
+                      toCPPTypes(ListItem(type(lvRefType(type(simpleType(T1))))) ListItem(type(simpleType(T1)))),
```
dwightguth: I would prefer you just write it as a CPPTypes, ie, as

`type(lvRefType(utype(T1)), utype(T1))`

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+          requires simpleType(T1) ==K simpleType(T2)
+*/
+
+
+     syntax Map ::= builtinNonvolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinVolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinNonvolatileArithAssignOperator(Expr, CPPType, ValueCategory, Expr, CPPType, ValueCategory) [function]
+
+     rule builtinNonvolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
```
dwightguth: `type(simpleType(T1))` is a bad idiom. If you want the unqualified version of a type, you should use utype.

dwightguth: I would recommend importing CPP-BUILTIN-SYNTAX and using functions from there if you need to construct elaborate types manually.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+                      noMethod, CPPLinkage, exSet(.Set)))
+                |-> builtinOp(E1 := E2))
+
+          requires isCPPEnumType(T1) andBool notBool isVolatileType(T1) andBool viableType(type(simpleType(T1)), T2, VC)
+
+     rule builtinNonvolatileEnumAssignOperator(...) => .Map [owise]
+
+     rule builtinVolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
+                      toCPPTypes(ListItem(type(lvRefType(volatile(simpleType(T1))))) ListItem(type(simpleType(T1)))),
```
dwightguth: similarly, try to avoid using simpleType here. If you need a const, non-volatile version of a type, write a function that does exactly that. it will prevent you from having to touch the modifiers, which will be a big help in refactoring the semantics if we decide to ever use the type modifiers in C++.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+                   (type(functionType(type(simpleType(T1)),
+                             toCPPTypes(ListItem(type(lvRefType(type(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                             noMethod, CPPLinkage, exSet(.Set))))
+               |-> builtinOp(E1 := E2)
+          requires isCPPArithmeticType(type(simpleType(T1))) andBool viableType(type(lvRefType(type(simpleType(T1)))), T1, VC1) andBool viableType(type(simpleType(T1)), T2, VC2)
+
+     rule builtinNonvolatileArithAssignOperator(...) => .Map [owise]
+
+     // There shouldn't be any user overload with the same parameter list for non-class T1.
+     rule addBuiltinCandidates(operator=, E1::Expr, T1:CPPType, VC1:ValueCategory, E2::Expr, T2:CPPType, VC2:ValueCategory, C1:CandidateSet, C2:CandidateSet)
```
dwightguth: you still need to check that T1 and T2 aren't class types if you want to do this.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+     syntax Map ::= builtinVolatileEnumAssignOperator(Expr, CPPType, Expr, CPPType, ValueCategory) [function]
+     syntax Map ::= builtinNonvolatileArithAssignOperator(Expr, CPPType, ValueCategory, Expr, CPPType, ValueCategory) [function]
+
+     rule builtinNonvolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
+                      toCPPTypes(ListItem(type(lvRefType(type(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                      noMethod, CPPLinkage, exSet(.Set)))
+                |-> builtinOp(E1 := E2))
+
+          requires isCPPEnumType(T1) andBool notBool isVolatileType(T1) andBool viableType(type(simpleType(T1)), T2, VC)
```
dwightguth: you want to check only for isCPPEnumType and leave the rest for overloading to resolve.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+
+     rule builtinNonvolatileEnumAssignOperator(...) => .Map [owise]
+
+     rule builtinVolatileEnumAssignOperator(E1::Expr, T1::CPPType, E2::Expr, T2::CPPType, VC::ValueCategory) =>
+                (type(functionType(type(simpleType(T1)),
+                      toCPPTypes(ListItem(type(lvRefType(volatile(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                      noMethod, CPPLinkage, exSet(.Set)))
+                |-> builtinOp(E1 := E2))
+
+          requires isCPPEnumType(T1) andBool isVolatileType(T1) andBool viableType(type(simpleType(T1)), T2, VC)
```
dwightguth: same with the side condition here.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+
+     rule builtinVolatileEnumAssignOperator(...) => .Map [owise]
+
+     rule builtinNonvolatileArithAssignOperator(E1::Expr, T1::CPPType, VC1::ValueCategory, E2::Expr, T2::CPPType, VC2::ValueCategory) =>
```
dwightguth: this is wrong. as I said, in the case of arithmetic types, pretty much all integer types can be converted to from an enumeration type. Since the builtin operators are finite since there is a finite number of arithmetic types, you need to exhaustively enumerate the candidates as a constant of sort Map and add that constant to the mix and let overload resolution deal with it.

semantics/cpp14/language/translation/name.k
```diff
@@ -497,6 +497,76 @@ module CPP-TRANSLATION-NAME
...
+                   (type(functionType(type(simpleType(T1)),
+                             toCPPTypes(ListItem(type(lvRefType(type(simpleType(T1))))) ListItem(type(simpleType(T1)))),
+                             noMethod, CPPLinkage, exSet(.Set))))
+               |-> builtinOp(E1 := E2)
+          requires isCPPArithmeticType(type(simpleType(T1))) andBool viableType(type(lvRefType(type(simpleType(T1)))), T1, VC1) andBool viableType(type(simpleType(T1)), T2, VC2)
+
+     rule builtinNonvolatileArithAssignOperator(...) => .Map [owise]
+
+     // There shouldn't be any user overload with the same parameter list for non-class T1.
```
dwightguth: this is wrong; you need to actually perform this filtering. This is easy however, you just need to do a map update operation in which any key in the builtin candidate set also present in the non-member candidate set is replaced.

semantics/cpp14/language/translation/overloading.k
```diff
@@ -18,6 +18,8 @@ module CPP-OVERLOADING-SYNTAX
 
      syntax BuiltinOp ::= builtinOp(Expr)
 
+     syntax Bool ::= viableType(CPPType, CPPType, ValueCategory) [function]
```
dwightguth: this should be unnecessary if you follow my other comments.

semantics/cpp14/language/translation/name.k
h0nzZik: I am not sure with the non-volatility requirement. I would think we should keep it, because non-volatile argument is not convertible to volatile type (and 13.3.1.2:3.3.3).

dwightguth: I mean, yes, the semantics states that only those builtin candidates for which types are convertible are part of the candidate list, so technically that's correct as far as it goes. But in terms of the actual semantics of overload resolution it doesn't actually matter whether we check that at this point in time because if we don't, it will get checked automatically when we compute the set of viable candidates. As far as I can tell that provision is there simply to make the candidate set finite in size.

The purpose of the side condition on this rule isn't to decide whether or not a conversion can occur, it's to determine whether or not `T1 operator=(T1&, T1)` is part of the list of builtin candidates for operator= in 13.6.

I would prefer to keep it that way because it separates concerns much better.

dwightguth:
I'm sorry, unfortunately I'm going to have to ask you to stop working on this because my own critical path has led me to this exact same issue (builtin candidates for relational operators on enumerations), and I can't afford to take the time that would be needed in order to wait for you to create something that passes through code review.

If you want something to work on that most likely will not overlap with what I'm working on, at least for the time being, I would suggest tackling the issue of zero-initialization of class types.

Essentially what we want here is for the zeroInit expression for class types to evaluate to a compound initialization in which the first operation writes a zero byte to all the bytes of the class iff the object, when evaluated as part of executing the initializer, is found to have automatic, allocated, or dynamic storage (and otherwise does nothing), and the second operation returns an lvalue of the object being initialized.

h0nzZik:
Ok, I will look at it. Later I would like to work on `constexpr`.

dwightguth:
Hmmm. Yeah, constexpr would be a good second choice probably. Essentially what you'll need to do there is make it so that constexpr functions evaluate during the translation semantics.

h0nzZik:
Regarding the zero initialization: why should zeroInit not initialize class variables of static storage duration?

dwightguth:
Because those objects are already implicitly zero-initialized by the memory reading mechanism (cf. getUninitializedBits). This saves us significantly in the size of the executable of programs with very large static arrays.

Note that the same is true of thread-local storage.

PR #279, opened 2017-04-19
Clang-Kast: What if presumed location is not valid?

h0nzZik:
`SourceManager::getPresumedLoc()` may return invalid location, which actually happens on my system.


dwightguth:
This has been fixed in the latest master.

PR #280, opened 2017-04-26
Constexpr for global variables

h0nzZik:
Reading global constexpr variable is now a compile-time expression.

Unfortunately, this duplicates some code from `cpp14/language/execution/io.k`,  (`errorReadOutOfBounds`, `getUninitializedBits` etc). Where to put the common code?


tests/unit-pass/global-constexpr.C
```diff
@@ -0,0 +1,3 @@
+constexpr int a = 1;
+constexpr int b = ::a;
```
dwightguth: this test isn't sufficient to exercise this code because it should have already been passing, just with b being dynamically initialized at program startup. I would recommend the following pattern for testing that constexpr is working as intended:
```
#include<stdlib.h>

constexpr int a = 1;
const extern int b;
int c = b;
constexpr int b = a;

int main() {
  if (c != 1) abort();
}
```

Here the test passes if and only if c is initialized after b, which occurs only if b is initialized statically and c is initialized dynamically.

h0nzZik:
I have fixed the mentioned deficits some time ago, but forgot to announce it in this discussion.

h0nzZik:
@dwightguth I have cleaned the imports, as you said in #292

PR #286, opened 2017-08-05
fix EAgain

h0nzZik:


dwightguth:
Fixed in 2f52176ed027188188309f2350f5790b11a361a3

PR #292, opened 2017-10-03
Fix base class access

h0nzZik:
Converting class instances to their base class as well as accessing an element of base class worked correctly only if:
- the class had only one base class.
- if the base class had no virtual function, then the derived class had no virtual function.

(I've discovered this when working on zero initialization of classes.)

### DIFF
--- a/semantics/cpp14/language/common/typing.k
+++ b/semantics/cpp14/language/common/typing.k
@@ -277,6 +277,8 @@ module CPP-TYPING-SYNTAX
 
      syntax CPPType ::= implicitObjectParameterType(CPPFunctionType) [function]
 
+     syntax LVal ::= getBaseClassSubobject(base: LVal, baseType: CPPClassType) [strict]
+
      syntax Bool ::= isBaseClassOf(base: CPPClassType, derived: CPPClassType) [function]
                    | isVirtualBaseClassOf(CPPClassType, CPPClassType) [function]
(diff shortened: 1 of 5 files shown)


semantics/cpp14/language/common/typing.k
```diff
@@ -277,6 +278,8 @@ module CPP-TYPING-SYNTAX
 
      syntax CPPType ::= implicitObjectParameterType(CPPFunctionType) [function]
 
+     syntax LVal ::= getBaseClassSubobject(base: LVal, baseType: CPPClassType) [strict]
```
dwightguth: you're making this a constructor for the LVal sort, which is likely to cause huge problems. The only constructors for LVal should be le and lv.

h0nzZik:
Fixed

dwightguth:
I just noticed something else: we try to avoid importing a syntax module into another syntax module unless there is no equivalent module to import containing just the sorts. Can you update your c++ prs to reflect this and then tag me again? Then I will merge this.

h0nzZik:
@dwightguth For this PR it is done. PR #298 does not have this problem. I will update PR #280 later (after resolving its conflicts).

h0nzZik:
@dwightguth I think that the sorts/syntax/implementation distinction is closely related to [this talk](https://www.youtube.com/watch?v=su9ittf-ozk)

h0nzZik:
Moved there: https://github.com/runtimeverification/rv-match/pull/611

PR #298, opened 2017-10-15
Zero initialization of classes

h0nzZik:
Implements zero initialization of classes (and its members, base classes and padding), and arrays.

Depends on https://github.com/kframework/c-semantics/pull/292.


h0nzZik:
Moved here https://github.com/runtimeverification/rv-match/pull/612

PR #344, opened 2018-11-29
Reformatting C++ semantics

h0nzZik:
Note to self: we may easily fix the indentation, too.
```
let desiredIndentation = 5;
for b in bubbles; do
  let x = indentationOf(lines(b)[0])
  for l in lines(b); do
    let c = indentationOf(l);
    stripIndentation(l);
    addIndentation(l, desiredIndentation + max(0, c - x));
  done
done
```


chathhorn:
I'm pretty ambivalent about all the extra white space (I kinda prefer rules rewriting the same production, especially functions, to be grouped together without extra whitespace), but I'm fine with it if you like this style.

h0nzZik:
Yes, I would prefer short rules to be formatted like
```
rule foo(1) => 2
rule foo(2) => -1
```
but the readability does not suffer that much when they are like
```
rule foo(1) => 2

rule foo(2) => -1
```
.  But the readability definitely suffers from long rules being smashed together, so I started with the 'empty line after each rule' style.

PR #357, opened 2018-12-10
Cleanup Makefiles

h0nzZik:


h0nzZik:
@chathhorn Please review as if it was squashed. I will not squash it since I do not want to run the tests again.

h0nzZik:
I will do it today. I also want rv-match to stop using some target names (like `translation-semantics`) and use the filename-based target names instead: `$(file)`-semantics for `$(file).k`, e.g. `c11-translation-semantics`. I will do that in the same PR.

PR #365, opened 2019-01-17
Update 2-buffer-overflow.c: improving this test bench

mparisi20:
This probably doesn't make much of a difference, but I believe I've fixed two issues in this test bench that prevented them from testing what they were trying to test.  I've modified `dynamic_buffer_overrun_028()` and `dynamic_buffer_overrun_018()` so that the buffer overflows encountered when running these functions are the ones implied by their respective comment descriptions.

  * The problem I found with _018 is that, since `index` was outside the bounds of the `indexes` array, a buffer overflow occurs in this array before the value from it is used to access the `buf` array.

  * The problem I found with _028 is that, since `buf1+5` was outside the bounds of the dynamically-allocated `buf1` array, a buffer overflow occurs in this array before the value from it is used to access the `buf2` array.

I think these two changes would improve the test bench by having them clearly check for the exact situation that their comments describe.


chathhorn:
Jenk

chathhorn:
Jenkins: ok to test

chathhorn:
This looks good, thanks! I'll merge it when our CI gets unbroken.

mparisi20:
Thanks, glad to help!

PR #370, opened 2019-02-02
Added aliases in namespace and class scope, and a test for both

mickyabir:


chathhorn:
Jenkins: test this please

PR #377, opened 2019-04-05
Clang type inference

chathhorn:
- Passes template instantiations & type of `auto` decls deduced by clang to the C++ translation semantics.
- Removes template deduction and the `CPPTypeExpr` sort from the C++ semantics.


chathhorn:
@dwightguth please review when you get the chance.

dwightguth:
I didn't look in depth at the changes to the semantics, so you should still get someone else to review, but this looks good from my perspective.

chathhorn:
@h0nzZik please review when you get the chance.

h0nzZik:
Do we want to remove all the code? What if we will want to have semantics for templates in the future?

PR #378, opened 2019-04-08
Generate kore from c translation semantics

dwightguth:
Parses, compiles, emits kore, and validates the C translation semantics. Does not yet work with linking semantics, c++ translation semantics, or execution semantics. Does not yet successfully generate a decision tree.


semantics/cpp/language/translation/decl/class.k
```diff
@@ -546,8 +546,10 @@ syntax Stmt ::= Defaulted(SpecialMemberFunction, Class) [klabel(DefaultedSMF)]
           <initializers>... X |-> (T |-> I::Init) ...</initializers>
           requires isDeletedDefaultConstructor(X, T, C, FuncT, VM, I)
 
+     syntax KItem ::= "dotInit"
```
h0nzZik: Why we cannot have `.Init`?

dwightguth: Good suggestion, it just didn't occur to me. I'll change it.

semantics/common/builtin.k
```diff
@@ -15,10 +15,11 @@ module COMMON-BUILTIN
      imports K-REFLECTION
      imports C-TYPING-SYNTAX
 
-     syntax Map ::= #nativeDefinedBuiltins() [function, hook(C_SEMANTICS.nativeFunctions), impure]
+     syntax Map ::= #nativeDefinedBuiltins() [function, hook(C_SEMANTICS.nativeFunctions)]
      rule nativeDefinedBuiltins => #nativeDefinedBuiltins()
 
      // defined as a K rule for if the hook is not defined
-     rule #nativeDefinedBuiltins() => .Map
+     rule [[ #nativeDefinedBuiltins() => .Map ]]
```
chathhorn: What's the meaning of the `[[` `]]` here?

semantics/common/compat.k
```diff
@@ -81,8 +81,81 @@ module COMPAT-SYNTAX
 
 endmodule
 
+module COMPAT-KAST [kast]
```
chathhorn: Also the `[kast]` here?

semantics/common/builtin.k
dwightguth: This says that #nativeDefinedBuiltins() rewrites to .Map if we match the pattern after it (in this case it will always match because it's just matching on `<k> _ </k>`. The reason for this is to make it so that the entire configuration is passed as one of the arguments to the hook.

semantics/common/compat.k
dwightguth: This is so this module is not imported on the llvm backend.

semantics/c/language/translation/typing/expr.k
```diff
@@ -323,23 +308,22 @@ module C-TYPING-EXPR
...
+        #Or ((L <<= R)
+        #Or ((L >>= R)
+        #Or ((L &= R)
+        #Or ((L ^= R)
+        #Or ((L |= R)))))))))))) [macro]
 
      // assignment expression
-     rule typeof(Lbl:KLabel(t(_, Mods::Set, T::SimpleType),_) => t(noQuals, Mods, T))
-          requires isAssignmentKLabel(#klabel(Lbl))
+     rule typeof(assignmentOp(t(_, Mods::Set, T::SimpleType),_) => t(noQuals, Mods, T))
```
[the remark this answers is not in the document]
dwightguth: I'd rather we made a change like this separately. I'd also rather we get all of my existing code that I have in my branches merged before making such an involved refactoring. But if you want, I can add this to my list of things to do once I get the linking semantics to parse with llvm.

PR #407, opened 2019-05-16
Makefiles set KCC, KXX, CC and CXX only if not set

charala1:


charala1:
Prevents these `Makefile`s from overriding externally-supplied values for `KCC`, `KXX`, `CC` and `CXX`.

PR #416, opened 2019-05-22
Updated Makefile (variable recipe)

charala1:


charala1:
Obsolete as of https://github.com/kframework/c-semantics/pull/441 and https://github.com/kframework/c-semantics/pull/443

PR #420, opened 2019-05-24
Benchmarking single-stage jenkinsfile

charala1:


charala1:
This was only here to trigger CI for benchmarking purposes.

Issue #429, opened 2019-05-28
out-of-source build for clang-tools


charala1:
* It is now built with `cd clang-tools && cmake . && make`
* No easy way for the main `clean` target to remove the `cmake`- generated stuff
* This causes problems with top-level variable values propagating and staying inside `cmake`- generated `Makefile`s. The problems manifest when re-building locally.

charala1:
Fixed in https://github.com/kframework/c-semantics/pull/433

PR #433, opened 2019-05-30
clang-tools build refactoring

charala1:
* Fixes https://github.com/kframework/c-semantics/issues/429
   - out-of-source build for `clang-tools`
   - `make clean` removes `clang-tools/build`
   -  Added `build/` to `clang-tools/.gitignore`

* Improved `clang-tools/CMakeLists.txt`.
  - Removed hard-coded library paths. Using
     `find_package` instead.
  - Modernized other code.
  - Changed minimum version to 3.5
  - Removed the installation targets
     - they are not maintained/used

* Removed some compiler warnings for `clang-tools/*.cc`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ pipeline {
     stage('Build') {
       steps {
         sh '''
-          eval $(opam config env)
+          eval "$(opam config env)"
           make -j4
         '''
       }
@@ -30,7 +30,7 @@ pipeline {
     stage('Test') {
       steps {
         sh '''
-          eval $(opam config env)
+          eval "$(opam config env)"
           make -C tests/unit-pass -j$(nproc) os-comparison
         '''
       }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ pipeline {
     stage('Build') {
       steps {
         sh '''
-          eval "$(opam config env)"
+          eval $(opam config env)
           make -j4
         '''
       }
@@ -30,7 +30,7 @@ pipeline {
     stage('Test') {
       steps {
         sh '''
-          eval "$(opam config env)"
+          eval $(opam config env)
           make -C tests/unit-pass -j$(nproc) os-comparison
         '''
       }
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ CXXFLAGS := -std=c++17
 
 K_DIST := $(realpath $(K_BIN)/..)
 
+CLANG_TOOLS_BUILD_DIR := clang-tools/build
+CLANG_TOOLS_BIN := $(CLANG_TOOLS_BUILD_DIR)/bin
+
 FILES_TO_DIST := \
 	scripts/kcc \
 	scripts/k++ \
@@ -30,8 +33,8 @@ FILES_TO_DIST := \
 	scripts/program-runner \
 	scripts/histogram-csv \
 	parser/cparser \
-	clang-tools/clang-kast \
-	clang-tools/call-sites \
+	$(CLANG_TOOLS_BIN)/clang-kast \
+	$(CLANG_TOOLS_BIN)/call-sites \
 	scripts/cdecl-3.6/src/cdecl \
 	$(K_DIST) \
 	LICENSE \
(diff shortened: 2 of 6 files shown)


clang-tools/ClangKast.cc
```diff
@@ -2371,41 +2389,42 @@ class GetKASTVisitor
...
+      ATOMIC_BUILTIN(__atomic_fetch_or, "__atomic_fetch_or")
+      ATOMIC_BUILTIN(__atomic_fetch_xor, "__atomic_fetch_xor")
+      ATOMIC_BUILTIN(__atomic_fetch_nand, "__atomic_fetch_nand")
+      ATOMIC_BUILTIN(__atomic_add_fetch, "__atomic_add_fetch")
+      ATOMIC_BUILTIN(__atomic_sub_fetch, "__atomic_sub_fetch")
+      ATOMIC_BUILTIN(__atomic_and_fetch, "__atomic_and_fetch")
+      ATOMIC_BUILTIN(__atomic_or_fetch, "__atomic_or_fetch")
+      ATOMIC_BUILTIN(__atomic_xor_fetch, "__atomic_xor_fetch")
+      ATOMIC_BUILTIN(__atomic_nand_fetch, "__atomic_nand_fetch")
+      #undef ATOMIC_BUILTIN
```
charala1: This `#undef` is the only thing I added in this function -- the rest shows different because I indented them.

Jenkinsfile
```diff
@@ -22,15 +22,15 @@ pipeline {
     stage('Build') {
       steps {
         sh '''
-          eval $(opam config env)
+          eval "$(opam config env)"
```
dwightguth: not sure this is actually correct, given that stdout of opam config env may contain whitespace.

Makefile
```diff
@@ -227,15 +230,12 @@ parser/cparser:
...
-.PHONY: clang-tools/clang-kast
-clang-tools/clang-kast: clang-tools/Makefile
-	@echo "Building the C++ parser..."
-	@$(MAKE) -C clang-tools
+$(CLANG_TOOLS_BIN)/%: $(CLANG_TOOLS_BUILD_DIR)/Makefile
+	@$(MAKE) -C $(CLANG_TOOLS_BUILD_DIR) $*
 
-clang-tools/Makefile:
-	@cd clang-tools && cmake .
+$(CLANG_TOOLS_BUILD_DIR)/Makefile:
```
dwightguth: this needs to either be a phony target or depend on CMakeLists.txt.

clang-tools/CMakeLists.txt
```diff
@@ -1,23 +1,45 @@
-cmake_minimum_required(VERSION 2.8.4)
-project(clang-tools)
+cmake_minimum_required ( VERSION 3.10.2 )
```
dwightguth: this is way too new a version of cmake to make mandatory.

clang-tools/CallSites.cc
```diff
@@ -1,4 +1,13 @@
 #define _XOPEN_SOURCE 700
+
+#ifdef __GNUC__
+#pragma GCC diagnostic push
```
dwightguth: if we're going to do this, let's just pass -Wno-comment on the command line.

Jenkinsfile
charala1: It does contain whitespace, even in this case. However, the format of the instructions issued makes it so it doesn't matter. Take a look at the perl instructions for `local::lib`, where a similar instruction is used to set one's environment, and they enclose it in quotes: https://metacpan.org/pod/local::libhttps://metacpan.org/pod/local::lib

Makefile
charala1: I can't believe I missed that. Thanks.

clang-tools/CMakeLists.txt
charala1: Version 3.10.2 is what comes with ubuntu bionic, which is the only OS release we are testing the semantics on.

In general, `cmake` has been dramatically improved in these later versions, and I'd argue it's okay to demand this new a version (this is an easy to obtain tool, even without a package manager). I could, however, see your point if you told me that the version installed on debian stretch is 3.7.2.

clang-tools/CallSites.cc
charala1: The idea is to prevent these warnings only when they come from headers outside our control.

Please see that I've surrounded the `#include` directives with these, and nothing else.

Jenkinsfile
dwightguth: I don't really care what perl does. I care what opam tells you to do, and this is not it.

clang-tools/CMakeLists.txt
dwightguth: We definitely should support at least as far back as Ubuntu 16.04 which has CMake 3.5.

clang-tools/CallSites.cc
dwightguth: Alright, I can accept that.

Jenkinsfile
charala1: Alright, I'll revert this.

clang-tools/CMakeLists.txt
charala1: Sounds reasonable. I double-checked that I am not using anything that is not available in 3.5.

clang-tools/CallSites.cc
charala1: Just a comment: this does not get rid of the warnings just yet. The reason is that GCC 7 has a bug that causes it to ignore this `pragma`. However, our current package dependencies fetch a newer version of gcc anyway, and in a later PR I'll force these things to be compiled with that (or clang), at which stage the warnings will actually disappear.

PR #437, opened 2019-05-31
Makefile: removed .SECONDEXPANSION

charala1:


Makefile
```diff
@@ -54,7 +54,7 @@ LIBC_SO := dist/profiles/$(PROFILE)/lib/libc.so
 LIBSTDCXX_SO := dist/profiles/$(PROFILE)/lib/libstdc++.so
 
 define timestamp_of
-    dist/profiles/$(PROFILE)/$(1)-kompiled/$(1)-kompiled/timestamp
```
chathhorn: Doesn't this just cause everything to be unconditionally rebuilt? I think the timestamp is generated by `kompile`.

charala1: Yes, it does. Good catch!

I know how to fix this, but it's more involved than how much time I should devote on it right now. Basically a complete refactoring of the `Makefile` and `semantics/Makefile` is required.

I'll put this PR on the back burner for now, and switch to rule profiling. I'll revisit this at a later time.

charala1:
Relevant goals achieved in https://github.com/kframework/c-semantics/pull/441

PR #449, opened 2019-06-07
More misc fixes for llvm backend

dwightguth:


h0nzZik:
@dwightguth What does `priority` mean?

PR #451, opened 2019-06-07
Prevent error recovery translating std libs

chathhorn:


chathhorn:
@charala1 please review when you get the chance.

PR #454, opened 2019-06-10
remove TypeMap functionality from C semantics

dwightguth:
we utilize the ability to match map keys on function applications to bypass the need for this special logic.


dwightguth:
Closing. I'll reopen this one later.

PR #459, opened 2019-06-10
Remove last of explicit klabels

dwightguth:


semantics/common/compat.k
```diff
@@ -49,12 +50,14 @@ module COMPAT-SYNTAX
 
      syntax List ::= times(Int, K) [function]
 
-     syntax Set ::= filterSet(Set, K) [function]
-     syntax List ::= filterList(List, K) [function]
+     syntax Set ::= filterSet(Set, Predicate) [function]
+     syntax List ::= filterList(List, Predicate) [function]
+
+     syntax Bool ::= Predicate "(" K ")" [function, avoid]
```
h0nzZik: What does `[avoid]` mean?

dwightguth: it means that in the event of an ambiguity between this piece of syntax and another, prefer the other piece of syntax to this one. it's being used to faciliate disambiguation with the isCPPFunctionType Predicate in name.k.

PR #462, opened 2019-06-11
Makefile handling of environment variables

charala1:
Respect environment variables:
- KOMPILE_FLAGS
- KDEP_FLAGS
- K_INCLUDE_PATH (colon-separated list)
- K_OPTS

### DIFF
--- a/semantics/Makefile
+++ b/semantics/Makefile
@@ -1,13 +1,11 @@
-# NOTE: PROFILE_DIR is required to be passed from above.
-
 # Overridden in the top-level makefile.
 BUILD_DIR := $(CURDIR)/../dist/semantics
 
 # Used internally.
 OUTPUT_DIR := $(abspath $(BUILD_DIR))
 
-# Intended to be passed from above.
-PROFILE_DIR ?= $(realpath $(CURDIR)/../profiles/x86-gcc-limited-libc)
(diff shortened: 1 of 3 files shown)


semantics/Makefile
```diff
@@ -1,22 +1,29 @@
...
+# Appending to whatever the environment provided.
+KOMPILE_FLAGS += --backend ocaml
+KOMPILE_FLAGS += --non-strict
+KOMPILE_FLAGS += --smt none
 
-KOMPILE_FLAGS += --backend ocaml --non-strict --smt none
-KDEP_FLAGS += $(addprefix -I ,$(KDEP_INCLUDE_DIRS))
+# Generate a makefile list from a colon-separated one.
+# K_INCLUDE_PATH comes from the environment.
+K_INCLUDE_DIRS := $(shell perl -pe 's/:/ /g' <<<$(K_INCLUDE_PATH))
```
dwightguth: if we just need a search/replace, we shouldn't need perl, right? makefile has text functions for this.

charala1: Yeah, I figured I'd do it in the simplest way possible (because perl is a dependency of this project). Let me see what I can do.

charala1: Done.

PR #464, opened 2019-06-11
CI: produce and archive timelogs.csv

charala1:


Jenkinsfile
```diff
@@ -24,9 +24,12 @@ pipeline {
         sh '''
           eval $(opam config env)
           export KOMPILE_FLAGS=-O2
-          make -j4
+          make -j4 profile-rule-parsing
         '''
       }
+      post { always {
+        archiveArtifacts artifacts: 'dist/timelogs.d/timelogs.csv', allowEmptyArchive: false, onlyIfSuccessful: true
```
dwightguth: allowEmptyArchive is false by default and you can run it only on successful builds by saying `post { success {`, so you should be able to just say `archiveArtifacts 'dist/timelogs.d/timelogs.csv'`

charala1: Done.

PR #469, opened 2019-06-13
Remove duplicate arguments to kompile

charala1:


scripts/kompile-wrapper
```diff
@@ -0,0 +1,11 @@
...
+
+# Remove optimization flags.
+SANITIZED_ARGS="$(echo "$@" | perl -pe 's/\-O.\s*//g')"
+
+# Append optimization flags, at most once each.
+echo "$@" | grep --quiet '\-O2' && SANITIZED_ARGS="${SANITIZED_ARGS} -O2"
+echo "$@" | grep --quiet '\-O3' && SANITIZED_ARGS="${SANITIZED_ARGS} -O3"
+echo "$@" | grep --quiet '\-Og' && SANITIZED_ARGS="${SANITIZED_ARGS} -Og"
+
+"$(dirname "${0}")/k/bin/kompile" ${SANITIZED_ARGS}
```
chathhorn: This kinda seems like overkill to me -- can't we just avoid passing duplicate arguments?

charala1:
Some make targets need to append arguments to what’s already in the
variable, and doing this check at the calling sites would lead to a huge
mess.

This PR is a temporary measure, and I am having second thoughts about it. I
think I should close this and open a PR with `k` instead.

chathhorn:
Another option might be a separte `KOMPILE_OPT_LEVEL` variable that can be overridden instead of appended to.

charala1:
How would we avoid an option appearing in both `KOMPILE_OPT_LEVEL` and
`KOMPILE_FLAGS`?

chathhorn:
I guess by not ever doing that? I don't see why we have to worry about that case.

charala1:
Because `make` is not supposed to ignore user-supplied flags unless it's
absolutely necessary. Think `CFLAGS=-g make`...

chathhorn:
Unless I'm mistaken, I really don't think we need to support anything other than changing the optimization level, though.

charala1:
> Unless I'm mistaken, I really don't think we need to support anything other than changing the optimization level, though.

The actual solution to this problem would come from `kompile` not crashing when it sees duplicate flags. Much like `gcc -g -g -g source.cc` doesn't crash...

PR #473, opened 2019-06-17
deleted:    tests/unit-fail-compilation/kcc_config

charala1:


dwightguth:
can we put files with this name in .gitignore too?

charala1:
Done!

chathhorn:
This was added to prevent differences in error messages depending on whether that file was produced or not. It might not be needed anymore because I think we're chopping that line now anyway, but please verify locally -- these tests are run on rv-match, not in the c-sem pr job.

charala1:
> This was added to prevent differences in error messages depending on whether that file was produced or not. It might not be needed anymore because I think we're chopping that line now anyway, but please verify locally -- these tests are run on rv-match, not in the c-sem pr job.

@chathhorn what should I verify then? That the file is not produced, or something else?

charala1:
Closing in favor of issue https://github.com/kframework/c-semantics/issues/492

PR #477, opened 2019-06-18
Faster Build and Cleanup part 6

charala1:


semantics/Makefile
```diff
@@ -25,6 +25,10 @@ K_INCLUDE_DIRS += $(PROFILE_DIR)/semantics/cpp
 # Appending to whatever the environment provided.
 KDEP_FLAGS += $(addprefix -I ,$(K_INCLUDE_DIRS))
 KOMPILE_FLAGS += $(addprefix -I ,$(K_INCLUDE_DIRS))
+KOMPILE_FLAGS += --no-prelude
+KOMPILE_FLAGS += -w all
+KOMPILE_FLAGS += -v
+KOMPILE_FLAGS += --debug
```
charala1: These flags were all used by all timestamp targets, with the exception of c-nd-thread. Now it too will use the same set of flags.

semantics/Makefile
```diff
@@ -125,19 +126,17 @@ endif
...
-K_FILE_DEPS += $(shell find $(OUTPUT_DIR) -type f -name '*.k')
-K_FILE_DEPS += $(shell find $(PROFILE_DIR)/semantics -type f -name '*.k')
-
-$(BASIC_DEPENDS): $(GENERATED_MAKEFILE_DIR)/basic-%.dep.mk: $(K_FILE_DEPS) \
+$(BASIC_DEPENDS): $(GENERATED_MAKEFILE_DIR)/basic-%.dep.mk: %.k \
                                                             | $(GENERATED_MAKEFILE_DIR)
-	$(KDEP) $(KDEP_FLAGS) -d "$(OUTPUT_DIR)/$*-kompiled" -- $*.k > $@
+	$(KDEP) $(KDEP_FLAGS) -d "$(OUTPUT_DIR)/$*-kompiled" -- $*.k \
+		| perl -pe 's|(.*?)\s*:|\1 $@:|g' \
+	> $@
```
charala1: The perl expression forces `$@` to depend on the same things that the targets inside `$@` depend on.

chathhorn: But the dependencies themselves really need to be recalculated if any of the `.k` files change, right? I don't see how you can avoid that.

charala1: That's exactly what this does. It makes the generated makefiles depend on the same things that the targets inside depend on. In other words, if any of the targets inside is out of date, so is the generated makefile.

PR #485, opened 2019-06-19
Fix some unintended non-determinism.

chathhorn:
* removes some unintended non-determinism found via Traian's tool
  * enables more of the unit-pass tests on the OS semantics and fixes some issues turned up by that (significantly increasing the PR job time, but I think it's probably necessary for the short term at least)
  * other misc fixes that would cause failures in the Match CI job


chathhorn:
@dwightguth this is ready for real now when you get the chance.

tests/unit-pass/Makefile
```diff
@@ -1,18 +1,12 @@
...
-	./stdio-fwrite.c ./stdio-temptest.c ./compcertTSO-2.c \
-	./stdio-ferror.c ./stdio-fgets.c ./stdio-fseek.c ./stdio-scanf-flush.c ./stdio-ungetc.c \
-	./stdio-sscanf.c ./stdio-fileno.c \
-	./inst-check.c ./pthreads-starve.c ./fopen-enoent.c \
-	./oldstyle-main.c ./struct-ini-1.c ./j016.c ./include-time.C \
-	./new23.C ./new41.C
+OS_STDIO_EXCLUDES = $(wildcard ./stdio-*.c)
+OS_EXCLUDES = $(OS_STDIO_EXCLUDES) \
+	./pthreads-starve.c ./fopen-enoent.c ./oldstyle-main.c ./j016.c \
+	./struct-ini-1.c ./20051113-1.c ./20050502-1.c ./string-opt-5.c ./inst-check.c \
```
dwightguth: why did this list of tests change?

chathhorn: Those were all excluded before -- previously it was only running the `j*.c` tests. I just enabled a bunch (~800) more tests.

Some tests fail on the OS semantics (like the `stdio-*.c` ones) because of missing stdlib support, but others fail for reasons I'm not clear about. I've fixed several issues here, but this needs to be revisited.

PR #499, opened 2019-06-25
Using a custom logger for the build

charala1:
* Added scripts/build-logger.h
* modified: Makefile
* modified: semantics/Makefile


charala1:
Made part of https://github.com/kframework/c-semantics/pull/500

PR #500, opened 2019-06-25
More Makefile improvements

charala1:


Makefile
```diff
@@ -41,13 +41,7 @@ CXXFLAGS += -Wextra
 CXXFLAGS += -Werror
 CXXFLAGS += -pedantic
 
-# We export these so they are available to the
-# `clang-tools` target.
-export CC := $(PROFILE_DIR)/cc
-export CXX := $(PROFILE_DIR)/cxx
```
charala1: @dwightguth Drawing your attention to this, and the relevant portion later on. While these work okay for the semantics, they cause the match build to fail. The reason is that they are also picked up by targets such as clang-tools, which don't work with these values. As far as I can tell, the cc/cxx under $(PROFILE_DIR) are only required for some native targets.

Makefile
```diff
@@ -236,10 +229,12 @@ $(LIBC_SO): $(call timestamp_of,c-cpp-linking) \
 				-shared \
 				-o $(OUTPUT_DIR)/profiles/$(shell basename $(d))/lib/libc.so \
 				*.c $(KCCFLAGS) -I .)
-	$(info $(PROFILE): Done translating the C standard library.)
+	@$(LOGGER) "$(PROFILE): Done translating the C standard library."
 
 
 .PHONY: $(PROFILE)-native
+$(PROFILE)-native: CC := $(PROFILE_DIR)/cc
+$(PROFILE)-native: CXX := $(PROFILE_DIR)/cxx
```
charala1: This is the added code relevant to https://github.com/kframework/c-semantics/pull/500/files#r297394004

PR #518, opened 2019-07-03
BindReference in translation semantics

h0nzZik:
Refactorings that will enable us to move parts of `bindReference` computations (temporary materialization for now and conversions in future) from execution semantics to translation semantics. Mainly removes a call to `bindReference` from execution semantics.


h0nzZik:
Closed in favor of https://github.com/kframework/c-semantics/pull/522.

PR #519, opened 2019-07-03
Dockerfile: upgrade jdk to v11

charala1:


charala1:
Obsolete as of https://github.com/kframework/c-semantics/pull/554.

PR #522, opened 2019-07-04
BindReference in translation semantics #2

h0nzZik:


h0nzZik:
I can reproduce it localy (using `--reverse-rules`). The left argument to `bindReference` is of non-reference type:
```
bindReference(
	lecpp(
		`_.___CPP-SYNTAX`(
			`*__CPP-SYNTAX`(`This`(.KList)),
			`no-template_CPP-SYNTAX`(.KList),
			`Name`(classId(`GlobalNamespace`(.KList),classSpecifier(`Struct`(.KList),`Identifier`(#token("\"S\"","String")),`.List{"_,__CPP-DYNAMIC-OTHER-SORTS"}`(.KList))),`Identifier`(#token("\"b\"","String")))),
		hasTrace(`ExecName`(classId(`GlobalNamespace`(.KList),classSpecifier(`Struct`(.KList),`Identifier`(#token("\"S\"","String")),`.List{"_,__CPP-DYNAMIC-OTHER-SORTS"}`(.KList))),`Identifier`(#token("\"b\"","String")))),

		tcpp(quals(`.Set`(.KList)),`.Set`(.KList),`int_CPP-TYPING-SYNTAX`(.KList))
	),

	lecpp(`_.___CPP-SYNTAX`(`*__CPP-SYNTAX`(`This`(.KList)),`no-template_CPP-SYNTAX`(.KList),`Name`(classId(`GlobalNamespace`(.KList),classSpecifier(`Struct`(.KList),`Identifier`(#token("\"S\"","String")),`.List{"_,__CPP-DYNAMIC-OTHER-SORTS"}`(.KList))),`Identifier`(#token("\"a\"","String")))),hasTrace(`ExecName`(`NoNNS`(.KList),`Identifier`(#token("\"a\"","String")))),tcpp(quals(`.Set`(.KList)),`.Set`(.KList),`int_CPP-TYPING-SYNTAX`(.KList))))~>
`#freezer#SemanticCastToK0_54`(.KList)
```

This happens because of [these rules in dynamic.k](https://github.com/kframework/c-semantics/blob/master/semantics/cpp/language/common/dynamic.k#L633-L634):
```
rule le(E::Expr, Tr::Trace, cppRefType #as T::CPPType) => le(E, Tr, innerType(T))
rule xe(E::Expr, Tr::Trace, cppRefType #as T::CPPType) => xe(E, Tr, innerType(T))
```
Most likely, we need these rules, because `le`,`xe`, and `pre` of reference types are not `KResult`s  - at least not in the translation semantics, see [this rule](https://github.com/kframework/c-semantics/blob/master/semantics/cpp/language/translation/init.k#L20):
```
rule isKResult(le(_, _, T::CPPType) #Or xe(_, _, T::CPPType) #Or pre(_, _, T::CPPType)) => notBool isCPPRefType(T)
```
But why we need that rule?

h0nzZik:
I still have problems binding a reference to temporary. Do we already have some support for temporaries? The standard says:

> 12.2/5 [...] The temporary to which the reference is bound or the temporary that is the complete object of a subobject to which the reference is bound persists for the lifetime of the reference except:
> (5.1) — A temporary object bound to a reference parameter in a function call (5.2.2) persists until the completion of the full-expression containing the call.

Multiple cases may happen:
1. The reference have automatic storage duration.
2. The reference have static storage duration (e.g. it is a global or local static variable).
3. The reference is a function parameter.
4. The reference is bound to a subobject of a temporary object, e.g. a member of a class.

Also, I heard that C++17 changed some rules about temporaries to support guaranteed copy elision - does that affect reference binding?

This is an interesting corner case for binding reference to a temporary:
```
struct S {
    int const &t = 3;
};
// S s1; // fails in implicitly-defined default constructor
S s{4}; // ok
```

h0nzZik:
To create a temporary, `bindReference` may use [`DeclareObject` from translation/decl/declarator.k](https://github.com/kframework/c-semantics/blob/master/semantics/cpp/language/translation/decl/declarator.k#L12). However, `bindReference` would need to know the duration (`AutoStorage`/`StaticStorage`) of the reference (or the complete object) that is being created/initialized. `DeclareObject` has this information. How do we transfer it down to `bindReference`?

One option is to add a new parameter to `#figureInit`, `figureInit`, `aggInit`, `classAndUnionAggInit`, `classAggInit`, arrayInit,  //maybe `figureConstructorInit`, `figureConstructItem`, `figureConstruct`.
`listInit`

Another option is to add a new cell containing the duration of the object being created.

Example:
```cpp
struct S { int const &r; };
S s1[2] {3,4};
int main()
{
    S s2{4};
}
```

h0nzZik:
RV-Match test: https://github.com/runtimeverification/rv-match/pull/1071

PR #539, opened 2019-07-19
Dockerfile: using internal registry image

charala1:


Dockerfile
```diff
@@ -1,29 +1,4 @@
...
-    && apt-get install --yes locales \
-    && locale-gen --no-purge "${LOCALE}"
-
-ENV LANG="${LOCALE}" \
-    LC_ALL="${LOCALE}" \
-    LANGUAGE="en_US:en"
-
-RUN update-locale
-
+FROM 10.0.0.21:5201/ubuntu-rv:bionic
```
charala1: This image sets up the locale and the default user.

Jenkinsfile
```diff
@@ -1,64 +1,124 @@
...
+      agent any
+      stages {
+        stage ( 'Build docker image' ) { steps {
+          script {
+            img = docker.build "${IMG_RELEASE_TAG}"
+          }
+        } }
+        stage ( 'Push to dockerhub' ) { steps {
+          script {
+            docker.withRegistry( '', 'rvdockerhub' ) {
```
charala1: The second argument is the credentials id corresponding to the "dockerhub@runtimeverification.com" user.

PR #541, opened 2019-07-23
C++ to kore parser

chathhorn:


clang-tools/ClangKast/ClangKast.cc
```diff
@@ -0,0 +1,371 @@
...
+string Kast::KToken::toKString(bool b) {
+  return string(b ? "true" : "false");
+}
+
+string Kast::KToken::toKString(llvm::APInt i) {
+  return string(i.toString(10, false));
+}
+
+string Kast::KToken::toKString(unsigned i) {
+  char *buf = new char[11];
```
h0nzZik: This leaks memory. It would be better if you just allocated on stack:
```cpp
char buf[11];
```

clang-tools/ClangKast/ClangKast.cc
```diff
@@ -0,0 +1,371 @@
...
+  unsigned precision = f.semanticsPrecision(f.getSemantics());
+  unsigned numBytes = 6      // max length of signed short
+                    + 4      // -1.E
+                    + 3 + 11 // p4294967295x16
+                    + 1      // null byte
+                    + precision;
+  SmallVector<char, 80> buf;
+  f.toString(buf, 0, 0);
+  char *data = buf.data();
+  data[buf.size()] = 0;
```
h0nzZik: You write one element past the end.

clang-tools/ClangKast/ClangKast.h
```diff
@@ -0,0 +1,139 @@
...
+public:
+
+  // Add a Kast::Node to the KAST. The next nodes added should be the children (bredth-first).
+  template <class T>
+  static void add(const T & node);
+
+  // Print the KAST.
+  static void print();
+
+  class Node {
```
h0nzZik: This class should have a virtual destructor (see https://stackoverflow.com/a/461224/6209703)

PR #546, opened 2019-08-07
Fix functional cast

h0nzZik:
`#computeUDCTo` does not play well with constructors that do not have exactly one parameter -> filter them out.
RV-Match test: https://github.com/runtimeverification/rv-match/pull/1087


h0nzZik:
TODO: check cast with multiple parameters and add a test for it.

PR #552, opened 2019-08-12
Using kframework base image

charala1:
* FROM runtimeverificationinc/kframework:ubuntu-bionic
  - most APT packages installed from there
* Environment set in Makefile
* Pull perl installation from https://hub.docker.com/r/runtimeverificationinc/perl

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,8 @@ RUN     apt-get update -q \
           openjdk-8-jdk   \
           libstdc++6 
 
+# TODO (Minas)
+# We should not need jdk-8. The base image has jdk-11 installed.
 RUN update-alternatives --set java /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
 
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -41,8 +41,6 @@ pipeline {
           steps {
             script { img.inside {
               sh '''
-                eval $(opam config env)
-                eval $(perl -I "~/perl5/lib/perl5" -Mlocal::lib)
```
chathhorn: What's the reasoning for moving these out of the Jenkinsfile? The alternative here of duplicating these across multiple makefiles doesn't seem particularly optimal.

charala1: I've been meaning to do this for a while now; I also saw it in the evm semantics makefiles that I've been working with in the last few days and really liked it.

It's not the primary purpose of this PR though, so if you don't agree, I'll revert this change.

PR #556, opened 2019-08-14
Fix aggregate destruction

h0nzZik:


h0nzZik:
failed functional-cast-4.C

The configuration has `#computeUDCTo` at the top of the K cell. That is probably the error fixed in https://github.com/kframework/c-semantics/pull/546, so I need to shift the priority there.

h0nzZik:
When constructing an aggregate that contains members of class types, these members need to be temporarily added to `<locally-constructed>`, so that if the construction of the aggregate ends with an exception, the members that are already constructed get destructed. However, before adding the aggregate itself to `<locally-constructed>`, its members should be removed from there. Currently, I wrap the construction of members in `pushBlock` and `popBlock`; however, that is problematic for global variables. So maybe it would be better to just save and restore the `<locally-constructed>` cell.

h0nzZik:
RV-Match test: https://github.com/runtimeverification/rv-match/pull/1090

PR #558, opened 2019-08-28
LLVM backend support

chathhorn:


dwightguth:
Let's wait to merge this until we get all the llvm backend tests passing. Right now it seems to be disabling C++, which we probably don't want to do in master.

PR #564, opened 2019-09-02
Delete local object just after its destruction.

h0nzZik:
The cell `<local-addresses>` is no longer used
in the C++ semantics.

### DIFF
--- a/semantics/cpp/language/common/dynamic.k
+++ b/semantics/cpp/language/common/dynamic.k
@@ -574,8 +574,6 @@ module CPP-DYNAMIC-OTHER-SYNTAX
 
      syntax KItem ::= seqstrict(StrictList) [klabel(seqstrictcpp)]
 
-     syntax SymBase ::= baseOfLoc(SymLoc) [function]
-
 endmodule // CPP-DYNAMIC-OTHER-SYNTAX
 
 module CPP-DYNAMIC-OTHER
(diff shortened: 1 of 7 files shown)


h0nzZik:
The file `execution/decl/declarator.k` contains an error: if the initializer throws, the variable is not registered for destruction; the destructor is not called (correctly), but also the allocated memory (`allocObject`) is not deleted (via `deleteObject`). We may fix this by doing `registerForDestruction` before `initializeObjectExec`, but with the `IsConstructor` flag set; the flag would be cleared after `initializeObjectExec`.

h0nzZik:
The fix contains another bug: if `initializeObjectExec` adds something to`<locally-constructed>`, then `updateRegisteredForDestruction` changes the newly added item. Even if we fix this, the principle is wrong: if we do `registerForDestruction` with `IsConstructor` flag set before `initializeObjectExec`, if `initializeObjectExec` throws an exception, the object gets destructed (and should not). We may fix this by adding a new `Enabled` flag into the entry. Also, I believe that we may remove the `IsVirtual` flag.

semantics/cpp/language/common/dynamic.k
```diff
@@ -1076,6 +1077,8 @@ module CPP-DYNAMIC-OTHER
 
      rule idOf(_ :: S::ClassSpecifier) => idOf(S)
 
+     rule baseOfLoc(loc(Base::SymBase, _) #Or loc(Base::SymBase, _, _)) => Base
```
chathhorn: I think there's already a `base(SymLoc)` function for this in `common/symloc.k`.

PR #573, opened 2019-09-13
Make size_t => unsigned-long

h0nzZik:
On my machine, there is a problem with the `new41.C` test. The test uses the `__SIZE_TYPE__` macro, that evaluates to `unsigned long`. Having `size_t => unsigned` in the semantics results in having two overloads for `operator new`, the implicit one with an `unsigned` parameter and the user-defined one with `unsigned long`, and the implicit one gets selected.


h0nzZik:
Test: https://github.com/runtimeverification/rv-match/pull/1098

PR #577, opened 2019-09-17
Free memory when `throw` is interrupted by `throw`

h0nzZik:


h0nzZik:
When we merge this, I will prepare a PR for the rv-match repository.

h0nzZik:
Merged in https://github.com/kframework/c-semantics/pull/593.

PR #596, opened 2019-11-11
Clang-based C parser #2

h0nzZik:


XVilka:
Maybe it makes sense to target clang-9 from the beginning?

h0nzZik:
> Maybe it makes sense to target clang-9 from the beginning?

Unfortunately, Ubuntu 18.04 does not have clang-9 in repositories.

PR #597, opened 2019-11-14
Fix restrict

h0nzZik:
A fix for https://runtimeverification.com/support/match/issues/36.


h0nzZik:
The problem is that we can have a `basedOn` refering to an object that no longer exists in memory. Maybe we could have another solution when we would remove such references every time an object is removed from the memory. With the current way we are risking that over time, some pointer accumulates lots of references to objects that no longer exists.

PR #598, opened 2019-11-27
Fix git clone command for docker installation

AlexCid:
The git command to clone the project is wrong - it uses checkout via ssh instead of HTTP.  Thus, it will fail, except for the project maintainers I guess...


h0nzZik:
Thanks.

PR #606, opened 2020-02-17
Do not perform lvalue conversion on LHS of assign

h0nzZik:


h0nzZik:
This is not correct, I misunderstood the code. With this 'fix' we miss a lots of UBs.

PR #608, opened 2020-02-26
Clever linking

h0nzZik:
If a symbol is not indirectly used by main, we should not require it to be defined. Implemented only for C.

### DIFF
--- a/semantics/c/language/execution/init.k
+++ b/semantics/c/language/execution/init.k
@@ -10,7 +10,6 @@ module C-EXECUTION-INIT-SYNTAX
      syntax KItem ::= "initMainThread'" "(" String ")"
      syntax KItem ::= "callAtExit" | "callAtQuickExit"
      syntax RValue ::= "NullPointerConstant" [function]
-     syntax String ::= getEntryPoint(Set) [function]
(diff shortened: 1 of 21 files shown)


h0nzZik:
TODO support different entry points than `main` (the `kcc -e entry_point` switch).

h0nzZik:
It completely suppresses them.

dwightguth:
That's not really ideal. They are still undefined behavior that the program has, and we shouldn't just not report it.

h0nzZik:
Ok, should I make it print a warning?

virgil-serbanuta:
I tested it on two autoconf projects (`gmp 6.1.2` and `secp256k1`)  projects that had `configure` scripts that were not based on autoconf (`z3` and `klee-uclibc`).

There is no change in behaviour in both the `c-semantics` `kcc` and the `rv-match` one.

The `c-semantics` `kcc` failed for `gmp` (`mpn_lshift_com optimization` error) and `klee` (link error for `ncurses` symbol).

The `rv-match` `kcc` failed for `gmp` (the `checking how to define a 32-bit word` step).

I think this should be fine to merge.

PR #611, opened 2020-03-04
sync changes to llvm backend into master

dwightguth:


semantics/common/alignment.k
```diff
@@ -2,7 +2,7 @@ module ALIGNMENT-SYNTAX
      imports SYMLOC-SORTS
      imports INT-SYNTAX
 
-     syntax Int ::= getAlign(SymLoc) [function]
+     syntax Int ::= getAlign(SymLoc) [function, withConfig]
```
h0nzZik: What does `withConfig` do?

semantics/common/alignment.k
```diff
@@ -14,16 +14,24 @@ module ALIGNMENT
      imports C-SETTINGS-SYNTAX
 
      rule getAlign(NullPointer) => cfg:alignofMalloc
-     rule getAlign(Loc::SymLoc) => getAlign'(base(Loc), offset(Loc)) [owise]
-     syntax Int ::= "getAlign'" "(" SymBase "," Int ")" [function, withConfig]
+     rule getAlign(Loc::SymLoc) => getAlign'(base(Loc), offset(Loc))
+          requires notBool isLinkerLoc(Loc) andBool Loc =/=K NullPointer
```
h0nzZik: Is this rule necessary when we have the `[owise]` rule below?

PR #613, opened 2020-03-12
fix nested union bug

h0nzZik:
This fixes the union bug reported by Denso.


dwightguth:
Can you describe in english exactly what this change is doing?

h0nzZik:
Previously, when the outer union did not contain the given member (`j` for the accompanying test), the original rule did not apply - no rule for `union` applied, and therefore the `[owise]` rule applied, which was a problem. I relaxed restrictions on that rule such that in applied in that case, but kept the guard strong enough so that it cannot fire when the rule below (the one with calls to `getUnnamed`, which probably handles the case of anonymous union member) is enabled. Note that the first line of the modified guard:
```
(notBool isNoType(getFieldType(F, T)) orBool isNoType(findFieldType(F, T)))
```
is exactly a negation of the guard from the second rule:
```
isNoType(getFieldType(F, T)) andBool notBool isNoType(findFieldType(F, T))
```

PR #614, opened 2020-03-12
Warnings for references to undefined symbols not accessible from main().

virgil-serbanuta:


semantics/linking/c-resolution.k
```diff
@@ -33,43 +39,47 @@ module LINKING-C-RESOLUTION
...
           <external-types> Types:Map </external-types>
           requires (notBool (X in_keys(Defs)) orBool notBool (X in_keys(Types)))
                andBool isMangledName(X)
           [structural]
-     rule <k> (.K => setTranslationLoc(L) ~> EXT-UNDEF("TDR2",
-                    "No definition for symbol with external linkage: "
-                     +String showCId(X) +String "."))
-              ~> resolveCReference(X:CId)
-          ...</k>
+     rule <k>  (resolveCReference(X:CId, ErrorHandling:LinkErrorHandling) => setTranslationLoc(L)
```
h0nzZik: Using a single colon generates a run-time check for the sort, which I think is not necessary here. Try using `::`.

virgil-serbanuta: Fixed.

PR #627, opened 2020-05-08
Reduce the cparser memory usage

virgil-serbanuta:
The stats below are for a 72MB input file:

Before this change:
Memory used: 44,5G
Time: 2m 59s 

After this change:
Memory used: 6G
Time: 3m 22s

### DIFF
--- a/parser/kPrinter.ml
+++ b/parser/kPrinter.ml
@@ -12,8 +12,16 @@ type printer_state =
   ; buffer : Buffer.t
   }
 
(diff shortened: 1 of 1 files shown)


dwightguth:
Six gigabytes still seems high, but this is a massive improvement. I assume you tested that it generates the exact same AST?

virgil-serbanuta:
Yes, it's the same AST for the 72G file. I didn't test on other files.

PR #647, opened 2020-11-11
Different overflow warning code for explicit unsigned casts.

virgil-serbanuta:


dwightguth (approved):
It seems like there ought to still be room for improvement, but this is good enough.

PR #649, opened 2020-11-26
Add locations for expressions

virgil-serbanuta:


semantics/c/language/translation/init.k
```diff
@@ -44,7 +48,89 @@ module C-TRANSLATION-INIT
           => UnknownCabsLoc
      rule filterCabsLoc(L:CabsLoc) => L [owise]
 
-     rule ExpressionLoc(K:KItem, _) => K
+     rule ExpressionLoc(K:K, L::CabsLoc) => ExpLoc(K, L)
+     rule typeof(ExpLoc(K:K, L::CabsLoc) => K)
```
dwightguth, line 52: rules for typeof ought to be in translation/typing/expr.k with the rest of them

semantics/c/language/translation/init.k
```diff
@@ -44,7 +48,89 @@ module C-TRANSLATION-INIT
...
+     rule utype(ExpLoc(K:K, L::CabsLoc) => K)
```
dwightguth, line 53: These rules confuse me. As I understand these functions don't normally apply to expressions?

semantics/c/language/translation/init.k
```diff
@@ -44,7 +48,89 @@ module C-TRANSLATION-INIT
...
+     rule type(ExpLoc(K:K, L::CabsLoc) => K)
+
+     syntax KItem ::= #freezerExpLoc(CabsLoc)
+
+     rule
+          <k> ExpLoc(K:K, L::CabsLoc) => K ~> #freezerExpLoc(L) ... </k>
+          <elab>... .K => L </elab>
```
dwightguth, line 60: This doesn't seem to be using the elab cell correctly. It's supposed to contain the elaborated result, but instead you're just appending a bunch of CabsLocs. That's fine for statements and declarations because they don't nest inside one another, but that doesn't work for expressions. What you want to be doing here is keeping the annotations unchanged and then dealing with the rest at execution time. It looks like you based this PR on the new C semantics, but it might have been better to model it based on the c++ semantics

dwightguth (changes requested):
This seems more complex than it needs to be. See my last comment; it's best to follow the same structure for expression locations that was used in the c++ semantics.

dwightguth:
I don't understand how this design came about. What happened to following the same design as the c++ semantics?

virgil-serbanuta:
> I don't understand how this design came about. What happened to following the same design as the c++ semantics?

I think that the main differences are the new location cell, the push-pop behaviour, and no context for ExpLoc, right?

Assuming that this is your objection:

I started with some code that was fairly similar to C++ one, and this should still be visible. However, if I remember correctly, there were a few cases when the warning/error for an instruction pointed to the expression inside the instruction instead of the entire instruction. I might be wrong, but there might have been some cases when the warning/error for an expression pointed at a random place inside the expression instead of the start of the expression.

Since I don't remember any examples, I'll try to switch back to using the main location cell, without push and pop, and using a context for ExpLoc, to see what diffs does it generate in the tests.

virgil-serbanuta:
I'll fix the remaining issues without checking them one by one. I'm not sure which approach is better for the user, but the last example, with the error that disappears, is worrying, and I'm not sure what to do about it.

So far, the diffs between the two approaches are (first is my code, the second is following the the cpp design):

* Division by 0 errors: pointer to the start of the expression => pointer to the last argument for the cpp approach. In this particular case, I guess it's a win for the cpp approach. In general, errors for operators switch form pointing to the start of the expression to to pointing to the last evaluated operator.
  Examples: 
  * return 1/a[0][0];  -- pointer on 1 vs a
  * ~0; -- pointer on ~ vs 0
  * char* p = (char*)-1;  -- pointer on "(" vs "-"
* Errors involving function calls (unimplemented function, stack traces, argument errors): pointer to the start of the expression => pointer to the last evaluated argument
  Example:
  *   return x6(0, u.b, va, s.bff, va); -- pointer on return vs va
* Some bitfield errors - pointer at the start of the struct => pointer on the (last?) bitfield size
  Example:
  * struct foo { short x : 5; }; -- pointer on "struct" vs "5"
* Some errors disappear because the last evaluated value comes from a system file.
  Example:
  *   int x = ULLONG_MAX;  -- pointer on ULLONG_MAX vs no error at all.

dwightguth:
If you've managed to solve the issue where errors caused by standard header macros get suppressed because the location is treated as inside the header instead of the location where the macro is used, I think that's probably a compelling argument that your solution is an okay one. I will take another look at it if you restore that version.

Thanks for being patient with me and coming up with a good example of the differences between the approaches.

virgil-serbanuta:
I reverted (but I also kept a backup branch, just in case it's still needed).

You probably already know this, but, for the record, I think this is what happens here:

With the cpp approach:
* We start evaluating the larger expression. We put its location in the main location cell.
* We evaluate the (constant) macro value, and we put its location in the location cell.
* We (implicitly) cast the macro value, which fails, and we try to use the macro location, which suppresses the error.

With my approach:
* We start evaluating the larger expression. We push its location on the stack.
* We evaluate the (constant) macro value, and we push and pop its location from the stack.
* We cast the value, which produces an error. The stack location is the one for the enclosing expression, which works.

I guess that we could try to fix the cpp approach by marking the location of system header expressions that are contained directly in user-based expressions as also being user-based, or something close to that.

dwightguth:
This still feels a little messy, but it seems to work and has demonstrated advantages over what I originally proposed, so I guess it's okay.

dwightguth (approved):

PR #650, opened 2020-12-01
Add configuration variable for hardware addresses

virgil-serbanuta:


semantics/c/language/common/memory/writing.k
```diff
@@ -75,7 +75,7 @@ module C-MEMORY-WRITING
           [structural]
      // Write global, shared data, or "allocated" duration.
      rule write(Loc:SymLoc, V:CValue, T:Type) => writeSpecific(Loc, V, T, false)
-          requires (isStaticDuration(Loc) orBool isAllocatedDuration(Loc))
+          requires (isStaticDuration(Loc) orBool isAllocatedDuration(Loc) orBool isHardwareLoc(Loc))
```
dwightguth, line 78: this seems weird. wouldn't all hardware locations be static storage? thus, you ought to just modify the isStaticDuration function.

virgil-serbanuta: Done, I think

dwightguth (approved):

PR #658, opened 2021-12-24
Update link to Ellison's thesis.

jcguu95:
The old link is dead.


ellisonch (changes requested):
The link you propose changing it to isn't my dissertation.  The correct link should be http://hdl.handle.net/2142/34297

It looks like the link to the POPL paper is also broken.  I think that should be https://fsl.cs.illinois.edu/publications/ellison-rosu-2012-popl.pdf

ellisonch:
Closing.  Have submitted more comprehensive fix in #659

